PR #5, opened 2020-10-13
Fix for instantiation of extensibility handlers

patrikhellgren:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Fix for instantiating extensibility handlers.
The previous code assumed that the extensibility handler assembly was located in the same folder as the executing assembly and that the handler assembly name specified in the HandlerType attribute for the Provider in the provisioning template was in its short form. If it was specified as a fully qualified name it would look for a file named something like "TestExtensibilityHandler, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null.xml". We would also not be able to load extensibility handlers from somewhere else than the same folder as the executing assembly.
With this fix handlers located in the same folder as the executing assembly will be loaded and we can both pre-load assemblies and use the AssemblyLoadContext.Default.Resolving event to load the handlers. It has been tested with all of them.

Credits to @pschaeflein who made the original PR [https://github.com/pnp/PnP-Sites-Core/pull/2764](https://github.com/pnp/PnP-Sites-Core/pull/2764). I just wanted this to get in the preview as soon as possible as we are in a big need for this.

### DIFF
--- a/src/lib/PnP.Framework/Provisioning/Extensibility/ExtensibilityManager.cs
+++ b/src/lib/PnP.Framework/Provisioning/Extensibility/ExtensibilityManager.cs
@@ -5,8 +5,6 @@ using PnP.Framework.Provisioning.ObjectHandlers;
 using PnP.Framework.Provisioning.ObjectHandlers.TokenDefinitions;
 using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Runtime.Loader;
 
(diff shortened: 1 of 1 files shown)


jansenbe:
Thanks @patrikhellgren , I'll look into it

PR #9, opened 2020-10-18
Fix some warnings related to XML comments

jackpoz:
Fixes the following warnings related to XML comments:
- CS1572 "XML comment on 'construct' has a param tag for 'parameter', but there is no parameter by that name" https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs1572
- CS1587 "XML comment is not placed on a valid language element" https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs1587
- CS1570 "XML comment on 'construct' has badly formed XML — 'reason'" https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs1570

On top of these there are a lot of warnings about parameters missing in the XML comments. Is there any meaning in adding those without any actual comment in another PR or is it better to keep the warnings as reminder ?

### DIFF
--- a/src/lib/PnP.Framework/Modernization/Cache/CacheManager.cs
+++ b/src/lib/PnP.Framework/Modernization/Cache/CacheManager.cs
@@ -1221,7 +1221,6 @@ namespace PnP.Framework.Modernization.Cache
         /// Get Term Path by ID
         /// </summary>
         /// <param name="context"></param>
-        /// <param name="termData"></param>
         /// <returns></returns>
         public void StoreTermSetTerms(ClientContext context, Guid termSetId, Guid sourceSspId, bool isSP2010, bool isSourceTerm)
         {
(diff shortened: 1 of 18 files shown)


jansenbe:
Thanks @jackpoz , always good to see folks also spending time on this type of updates as proper docs is important (and we've been doing a bad job with that for PnP Framework, PnP Core SDK is much better at that)

jackpoz:
Thanks @jansenbe . About all the missing XML comments for function parameters, what would be better, to define the XML comment with an empty string or keep the warnings until we slowly handle them adding a meaningful comment to that ?

Do you have any preference on how many warnings to deal with each PR ? I try to keep it with a reasonable amount of changes, not too few, not too many (and there are 2'400 warnings 😮  )

jansenbe:
Let's gradually add them, would be great if you could help. Just create a pr whenever you have processed a batch of them, there's no min/max size on PR's for this as it is not about changing code.

PR #11, opened 2020-10-18
fix url encoding issue when writing href values to client_LocationBasedDefaults.html as part of SetDefaultColumnValuesImplementation()

jvossers:
SharePoint is expecting a specific type of URL Encoding for href values in client_LocationBasedDefaults.html (the file which SharePoint produces to store the "Column default settings" for a document library) as can be observed when inspecting the file in the /Forms directory for the library after configuring defaults through the SharePoint UI. The type of encoding used by SharePoint leaves slashes alone and replaces spaces with %20 rather than '+' signs.

This PR addresses a bug in pnp framework (inside the SetDefaultColumnValuesImplementation method) where System.Web.HttpUtility.UrlEncode(path) is used to encode href values, which does not result in the type of encoding expected by SharePoint.


jansenbe:
Nice work @jvossers , will merge today

PR #20, opened 2020-11-08
Add some missing XML comments

jackpoz:
This PR add some missing XML comments and fixes some typos found while going through the comments.

As there are 2400 warnings, I started with [CS1573](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs1573) ones, so methods already with XML comments but with some parameters missing.

I tried to check how the same parameters/properties were commented in the existing code to keep them similar and give a better user experience, i.e. contructor parameter and public property having the same comment.
In some cases I checked how the methods are used or ```git blame``` to get a better understanding of what each parameter does.

I might keep pushing new commits to this PR but feel free to merge it whenever you want 😃


jansenbe:
Thanks @jackpoz - another batch of warnings gone with improved inline comments!

PR #24, opened 2020-11-12
Fixed NullReferenceException when adding Webhook Subscription

martinewald:
Hi, I noticed a NullReferenceException when i added a Webhook Subscription and fixed it.

Kind Regards
Martin


jansenbe:
Thanks for this PR @martinewald , will process it shortly

PR #27, opened 2020-11-16
Feature: Added support to extract all teams

gautamdsheth:
Currently, we can only extract 100 teams, if there are more than 100 teams then it won't work. This PR adds support for extracting >100 teams via the provisioning engine.

Please let me know if I should port it back to PnP Sites core as well.


jansenbe:
Good improvement @gautamdsheth ! Folks working with Teams do work in SPO, hence they should switch to PnP Framework soon. No point in backporting this to PnP Sites Core.

PR #28, opened 2020-11-17
Replaced graph beta call with already existing private method with v1 graph call

patrikhellgren:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Just replaces a graph beta call to remove a teams tab with a call to the already existing method that uses a v1 graph call


jansenbe:
Thanks @patrikhellgren !

PR #32, opened 2020-11-22
Improvements - removed some extra checks + maybe telemetry fix

gautamdsheth:
This PR 
1) Removes some extra checks for checking server version in publishing pages since we will always be in SPO
2) Removes check for ZoneId property as well since we will always be in SPO 
3) Fixed the deprecated method for telemetry tracking, not sure if it fixes the DLL hell issue but worth a check :)


jansenbe:
Thanks for this PR @gautamdsheth , point 1 and 2 make a ton of sense, for point 3 I'll need to do cross checking when I'll merge this PR.

jansenbe:
Telemetry changes were fine, merged.

Issue #34, opened 2020-11-23
Provisioning fails with ClientId/ClientSecret authentication (AcsAppOnly) 


sebastianmattar:
Hi!

We are using authentication by ClientId/ClientSecret. This causes some problems when using site provisioning for some object types (e.g. "SiteHeader") as access token generation is not working in all cases. 
I am looking into fixing it and supplying a pull request.

jansenbe:
Thanks for reporting this @sebastianmattar , looking forward for more details and a possible fix

luizfef06:
@sebastianmattar I am getting a SocketException at  PnP.Framework.Utilities.TokenHelper.AcsMetadataParser.GetMetadataDocument(String realm). Is it the same issue you are facing?

luizfef06:
This is the stack for me:

System.Net.WebException: An invalid argument was supplied. An invalid argument was supplied.
 ---> System.Net.Http.HttpRequestException: An invalid argument was supplied.
 ---> System.Net.Sockets.SocketException (10022): An invalid argument was supplied.
   at System.Net.Sockets.Socket..ctor(AddressFamily addressFamily, SocketType socketType, ProtocolType protocolType)
   at System.Net.Sockets.DualSocketMultipleConnectAsync..ctor(SocketType socketType, ProtocolType protocolType)
   at System.Net.Sockets.Socket.ConnectAsync(SocketType socketType, ProtocolType protocolType, SocketAsyncEventArgs e)
   at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.ConnectAsync(HttpRequestMessage request, Boolean allowHttp2, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.CreateHttp11ConnectionAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.GetHttpConnectionAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.SendWithRetryAsync(HttpRequestMessage request, Boolean doRequestAuth, CancellationToken cancellationToken)
[...]

PR #41, opened 2020-12-07
Fix ACS token generation

sebastianmattar:
- Fix ACS access token generation  #34 
- Add null pointer check
- Fall back to template LCID for ClientSitePage if none is provided


jansenbe:
Nice work @sebastianmattar , will process this in the coming day

PR #45, opened 2020-12-11
Fix warnings

jackpoz:
Fix just a couple of warnings:
1. XML comment for a parameter that doesn't exist anymore
2. Suppressing a warning CS0168 "The variable 'var' is declared but never used" instead of another CS0618 "'member' is obsolete: 'text'" (it was just a typo in the CS#### number)

There are a few warnings about Guid object being compared to null, i.e.:
```csharp
public void CacheTermsFromTermStore(Guid sourceTermSetId, Guid targetTermSetId, Guid sourceSspId, bool isSP2010)
        {
            // Collect source terms
            if (sourceTermSetId != null && sourceTermSetId != Guid.Empty) <--CS8073
```
with message
>The result of the expression is always 'true' since a value of type 'Guid' is never equal to 'null' of type 'Guid?'

Would it be ok to remove the unnecessary null check ? I don't think changing the parameter type to Nullable<Guid> makes that much sense as Guid.Empty can be used already with the same meaning.

If it's ok to remove the Guid != null checks, please let me know and I will update this PR.


jansenbe:
Hey @jackpoz , thanks for this and yes feel free to remove the unnecessary checks

jackpoz:
Well it seems https://github.com/pnp/pnpframework/commit/1267f7db90c8a91b314ca36794bfdcea00301bce fixed the Guid != null warnings and the typo in warning suppression, so this PR is ready with just 1 single warning fix.

PR #50, opened 2020-12-13
Fix - sensitivity label + preferred data location missing changes

gautamdsheth:
Looks like during some merge , we missed the Sensitivity Label and Preferred data location changes. 
This PR fixes that.


jansenbe:
Nice find, thanks @gautamdsheth

PR #58, opened 2020-12-18
Fix for unhandled null reference exception

patrikhellgren:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Fix for unhandled null reference exception if you did not specify a hierarchy or template file connector when provisioning a template with localization files.


jansenbe:
Thanks for this fix @patrikhellgren

PR #59, opened 2020-12-23
Feature - added support for chunked uploading of files

gautamdsheth:
Feature - Added support for chunked file upload to address some timeout issues for large files which can take >30 mins to upload


jansenbe:
Good addition @gautamdsheth ,  makes sense to transparently switch to a chunked upload in case of uploading large files

jansenbe:
Hey @gautamdsheth , I'm not yet merging your code as I think the implementation should depend on the stream and not on the filename. You can use the stream.Lenght to determine the file size and read chunks from the stream for the upload (see https://github.com/pnp/pnpcore/blob/dev/src/sdk/PnP.Core/Model/SharePoint/Core/Internal/FileCollection.cs#L84 for the equivalent PnP Core SDK implementation). Would you be able to update your PR with this approach?

gautamdsheth:
@jansenbe - changed it as you suggested, hope its ok :)

PR #60, opened 2020-12-23
Enhance allow read old pnp files

czullu:
This Enhancement allows to read the PnP-Files created with OfficeDevPnP.Core by replacing the references while we read.


czullu:
okay - see that you solved this already

PR #63, opened 2020-12-29
Feature - added additional props for Team Site No Group

gautamdsheth:
Feature - added HubSiteId, SensitivityLabel and SiteDesignId properties for Team sites without M365 groups.


jansenbe:
Nice one, thanks again @gautamdsheth !

PR #68, opened 2021-01-01
fix reading certificate bytes

mbakhoff:
Code should check the return value of `certfile.Read(certificateBytes, 0, (int)certfile.Length)` because it may read less than `certfile.Length` bytes. This patch delegates the work to X509Certificate2. A good alternative would be to use `System.IO.File.ReadAllBytes(certificatePath)`.

I also think that disposing the certificate here is not correct. Isn't the disposed certificate later used by ConfidentialClientApplication?

### DIFF
--- a/src/lib/PnP.Framework/AuthenticationManager.cs
+++ b/src/lib/PnP.Framework/AuthenticationManager.cs
@@ -216,19 +216,14 @@ namespace PnP.Framework
             {
                 ConfidentialClientApplicationBuilder builder = null;
 
-                using (var certfile = System.IO.File.OpenRead(certificatePath))
+                using (var certificate = new X509Certificate2(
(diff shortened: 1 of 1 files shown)


jansenbe:
Looks good, will process this one in the coming days. Thanks for sharing this fix @mbakhoff

jansenbe:
I've had to rewrite your PR a bit as both the original as your code resulted in an error when the builder used the cert. The key is to not dispose the cert while the builder is still needed

mbakhoff:
Thanks. Looks like your change https://github.com/pnp/pnpframework/commit/19c0052ca493bb226737ec53c424ceba7da80c48 fixes the dispose issue. However doesn't it also revert my change so that the bad Read issue is back?

PR #72, opened 2021-01-05
Parse tokens in the SearchCenterUrl

YannickRe:
The parsing already happens, but the result isn't used. This PR seeks to fix that.


jansenbe:
Good fix, thanks for sharing it @YannickRe

PR #82, opened 2021-01-08
Fix a null reference exception in cases where ClientContextSettings are null.

### LINKED ISSUE #70: "Object reference not set to an instance of an object" when provisioning ClientSidePages
**Version**: >= 0.86.0-preview (currently trying 0.101.0-preview)

### Issue
`Object reference not set to an instance of an object.` when applying a template (with ClientSidePage) to a web, using following code:
```
web.ApplyProvisioningTemplate(provisioningTemplate, applyingInformation);
```
Same code & template work when using a nuget package 0.84.0 for example. Seems to be introduced when PnP.Core was used for handling ClientSidePages (not sure if the issue lies in this library or with them).
My template is very simple and only provisions 1 ClientSidePage.

**Output**:
```
01/02 - Initializing engine
02/02 - ClientSidePages
```

**Stacktrace**:
```
Object reference not set to an instance of an object. at
PnP.Framework.Provisioning.ObjectHandlers.SiteToTemplateConversion.ApplyRemoteTemplate(Web web, ProvisioningTemplate template, ProvisioningTemplateApplyingInformation provisioningInfo, Boolean calledFromHierarchy, TokenParser tokenParser) at
Microsoft.SharePoint.Client.WebExtensions.ApplyProvisioningTemplate(Web web, ProvisioningTemplate template, ProvisioningTemplateApplyingInformation applyingInformation)
```

Peculiar thing: it does work without a problem when using PnP.PowerShell (latest nightly build, including PnP.Core 0.101.0-preview).

### Steps to reproduce
- Create a modern Communication Site
- Apply the template with web.ApplyProvisioningTemplate (in code, as mentioned: PowerShell works). I had to upload as zip, so it needs renaming to pnp first.
- See it fail with the `Object reference not set` error

[SitePagesTemplate.zip](https://github.com/pnp/pnpframework/files/5765698/SitePagesTemplate.zip)

YannickRe:
Fix a null reference exception in cases where ClientContextSettings are null. Fixes #70

### DIFF
--- a/src/lib/PnP.Framework/PnPCoreSdkAuthenticationProvider.cs
+++ b/src/lib/PnP.Framework/PnPCoreSdkAuthenticationProvider.cs
@@ -92,7 +92,7 @@ namespace PnP.Framework
             get
             {
                 var contextSettings = clientContext.GetContextSettings();
-                if (contextSettings.Type == Utilities.Context.ClientContextType.Cookie)
+                if (contextSettings != null && contextSettings.Type == Utilities.Context.ClientContextType.Cookie)
                 {
                     if (cookieContainer == null)
                     {


jansenbe:
Good fix for the issue as we indeed can't assume that everyone uses the AuthenticationManager to create contexts

PR #96, opened 2021-01-14
Added token parsing of team displayname in log message

patrikhellgren:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Added token parsing of team displayname in log message which would otherwise cause an exception if the displayname contained a token.


jansenbe:
Good improvement.

PR #119, opened 2021-01-22
Refactor ACS token creation

sebastianmattar:
Alright, so this is my proposal on how to handle token generation in the feature. I tried to untangle the ACS token generation bit, moved it to a separate namespace.

- Refactor ACS token creation
- Add geneirc support for OAuthAuthenticationProvider
- Implement OAuthAuthenticationProvider for ACS token generation
- Make ACS token generation thread-safe

### DIFF
--- a/src/lib/PnP.Framework/AuthenticationManager.cs
+++ b/src/lib/PnP.Framework/AuthenticationManager.cs
@@ -1,7 +1,7 @@
 ﻿using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.Extensibility;
 using Microsoft.SharePoint.Client;
-using PnP.Framework.Diagnostics;
+using PnP.Framework.Utilities;
 using PnP.Framework.Utilities.Context;
 using System;
 using System.Linq;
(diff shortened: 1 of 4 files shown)


jansenbe:
Hi @sebastianmattar , I do see you've tried to refactor a lot of things...although I'm pretty sure this can work it introduces additional complexities and dependencies that are not needed to solve your root issue being thread safety of TokenHelper.cs. We've been discussing this PR and decided we can't merge it in this shape due to above mentioned concerns. 

Wouldn't it be possible to fix the root concern (thread safety of TokenHelper.cs) by changing TokenHelper.cs itself without doing all these refactorings...keep in mind that ACS is a legacy technology which will be replaced by alternative scoped Azure AD permissions this year.

sebastianmattar:
Alright, just pushed a version which focusses on the thread-safety

jansenbe:
Thanks @sebastianmattar , appreciate the effort. We currently don't have proper contribution guidelines for PnP Framework and not complete enough ones for PnP Core SDK: an area for us to improve so it becomes easier for contributors to understand how to create their PR's.

jansenbe:
Just checked this, really nice solution ~ much better than the old thread model for token invalidation. Thanks @sebastianmattar !

PR #122, opened 2021-01-25
ExecuteQueryImplementation handle "forcibly closed by the remote host"

czullu:
Issue: while testing create complex template site from Azure Function i got at different locations in the code "forcibly closed by the remote host" Error even if i set System.Net.ServicePointManager.SecurityProtocol = System.Net.SecurityProtocolType.Tls12 | System.Net.SecurityProtocolType.Tls11 | System.Net.SecurityProtocolType.Tls; in initialization of AZ-Func.
To make sure it's not my dev-env i created a VM in Azure and moved the dev-Env to a VM in Azure in same Datacenter then my dev-tenant is - that way i'm quite sure that this is not a networkissue.

My workaround makes the code go into retry on "forcibly closed by the remote host" as i did not find the root cause.


jansenbe:
@czullu : is this something that recently came up? What http status code do you get back...500? Does happen for all tenants or just specific ones? 

This seems to be more of a backend issue somehow

czullu:
Hi Bert
I use 3 tenants to run our tests - 2 based in West Europe and one based in Switzerland. Since we not yet in production with pnpframework, that's all I have.
But I would agree that this is a backend issue. For me it means any time it happens the whole Export Template Process would abort - so I had to find a workaround to continue testing.

jansenbe:
@czullu : do you still see this issue appearing?

czullu:
> @czullu : do you still see this issue appearing?

i run all my tests with a branch including that fix - so i can't really tell right now.  Let me add some logging for this error and check the logs in a couple of days.

czullu:
To give some update on my analysis in short: 
I have the connection failure in certain setups always within ApplyTemplate - usually somewhere  in step "List instances (List & standard Fields)" (always different List / Field).

In Details:
-I have the issue when running the code in VS 2019 in a Guest VM but not when i run it on same PC on the Host itself. It seems  like the Issue came with the update of Windows 10 to 19042.804 for Host & Guest. 
-For some strange reason the issue in Guest VM does not happen if i run the traffic through Fiddler on the Guest VM.
-Running same code in a Azure VM B4ms has no issue but on a VM D4 v3 it has.

My conclusion: It's not in the code, so i would close this.

Issue #117, opened 2021-01-22
Failed to set ContentType on ClientSidePage


czullu:
@jansenbe 
The Error while set ContentType on ClientSidePage happens in ObjectClientSidePages in here  

```
           // Update page content type
            bool isDirty = false;
            if (!string.IsNullOrEmpty(clientSidePage.ContentTypeID))
            {
                //page.PageListItem[ContentTypeIdField] = clientSidePage.ContentTypeID;
                //page.PageListItem.UpdateOverwriteVersion();
                //web.Context.Load(page.PageListItem);
                fileAfterSave.ListItemAllFields[ContentTypeIdField] = clientSidePage.ContentTypeID;
                fileAfterSave.ListItemAllFields.UpdateOverwriteVersion();
                web.Context.Load(fileAfterSave.ListItemAllFields);
                isDirty = true;
            }
```

Checking both values before assign shows that the page has the ContentTypeId from the Lib while the Template has it from the Web.
fileAfterSave.ListItemAllFields[ContentTypeIdField]
0x0101009D1CB255DA76424F860D91F20E6C411800BD421750B0C84F43BC71B37A21DA5118

clientSidePage.ContentTypeID;
0x0101009D1CB255DA76424F860D91F20E6C4118

Error i get as ServerException: "Additions to this Web site have been blocked.\n\nPlease contact the administrator to resolve this problem."

Work on a fix to get List-CT based on clientSidePage.ContentTypeID - then compare and if different update. If fix is ok - i will submit.

jansenbe:
@czullu : would be great if you can submit a fix for this.

jansenbe:
@czullu : you also get this error is the page was checked in...maybe something to cross check

czullu:
> @czullu : you also get this error is the page was checked in...maybe something to cross check
Yes the Page is checked-in - it happens on default home.aspx after site creation.

jansenbe:
Closing issue now that PR has been merged

PR #135, opened 2021-01-26
Fix 117 failed to set content type on client side page

czullu:
Get always the best match for the ContentTypeId from Template in the List and set ListItem to this Content Type.

Issue: https://github.com/pnp/pnpframework/issues/117


jansenbe:
Thanks @czullu !

Issue #142, opened 2021-01-29
Issue when trying to create a new TeamSiteNoGroup sitecollection


magarma:
Hi, 
I'm trying to create a sitecollection of type "TeamSiteNoGroup", but it's giving an exception...

Sample of template
```
<pnp:SiteCollections>
            <pnp:SiteCollection xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
                                xsi:type="pnp:TeamSiteNoGroup" 
                                ProvisioningId="TEST" 
                                Title="{parameter:SiteTitle}" 
                                Alias="{parameter:SiteAlias}" 
                                Url="{parameter:SiteCollectionUrl}"
                                IsPublic="true" 
                                IsHubSite="false" 
                                Owner="{parameter:UPNOwner}" 
                                Description="REPLACED_BY_CONNECT">
                <pnp:Templates>
                    <pnp:ProvisioningTemplateReference ID="Atlas.Test.ModernTeamSite.JustAList"/>
                </pnp:Templates>
            </pnp:SiteCollection>
        </pnp:SiteCollections>
```
It's seems that the method GetRequestPayload is already adding the key WebTemplateExtensionId and then code gives the exception...
An item with the same key has already been added. Key: WebTemplateExtensionId

![image](https://user-images.githubusercontent.com/8925463/106291012-661a4900-624b-11eb-9c7b-a0d26424dba2.png)

...
I'll send a PR for this.

Thanks

magarma:
PR #143 
Thanks

jansenbe:
Fix has been merged, thanks @magarma  for delivering a problem with a fix linked it...definitely helps speed up handling the issue list ;-)

PR #143, opened 2021-01-29
Updating WebTemplateExtensionId value in payload dictionary.

magarma:
PR for fixing issue #142 
Instead of adding the key-value, It updates the value for key WebTemplateExtensionId in the payload dictionary.
image


jansenbe:
Thanks for this fix @magarma :-)

PR #147, opened 2021-01-30
Fix some warnings

jackpoz:
- CA1032: Implement standard exception constructors
- CA1507: Use nameof in place of string
- CA1834: Use StringBuilder.Append(char) for single character string
- CA1802: Use Literals Where Appropriate
- CA1825: Avoid zero-length array allocations


jansenbe:
Thanks for the cleanup work @jackpoz , highly appreciated!

PR #152, opened 2021-02-02
Fix Escaped whiteSpace break JSON in NewDocumentTemplates

czullu:
Fix since RegEx.Escape in ListContentTypeIdToken.CreateToken does escape whiteSpace with \ it breaks the JSON in NewDocumentTemplates


jansenbe:
Good improvement! Thanks again for sharing this one @czullu

PR #153, opened 2021-02-02
fix because Modern List Creation creates CTType with 0x01[Parent1][Parent2] but Parent1 does not exist

czullu:
fix because Modern List Creation (Lists-Dialog) creates CTType with 0x01[Parent1][Parent2] but Parent1 does not exist so it's not resolved above and we try to create the builtin CT 0x01 found by export and fail to create


jansenbe:
@czullu : that was a surprising find...did do some digging but so far I've not been able to 'find' the intermediate CT definition.

PR #166, opened 2021-02-10
Added Get/Add EventReceiver method to SiteExtensions

bhishma:
Added methods to return and register an event receiver at the site collection level (similar to web level).
```
GetEventReceiverById
GetEventReceiverByName
AddRemoteEventReceiver
```

It would resolve https://github.com/pnp/powershell/issues/241


bhishma:
While this is not going to really solve my issue because none of the Remote Event Receivers are being triggered when registering using modern authentication, I think this change would make it easier for people to register RERs at the site collection level. Please let me know if it needs more work.

PR #189, opened 2021-02-23
Fix: Changed the way we check the teams tab app id

patrikhellgren:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Fix for checking the teams tab app id. The old teamsappid property way only exists in the beta endpoint and nowadays seems to return null for some (if not all) apps. The way it is done in this PR is the way the documentation says is the way to do it.


jansenbe:
Thanks @patrikhellgren !

PR #195, opened 2021-02-26
Allow to have Custom UserAgent when use PnPHttpClient.Instance.GetHttpClient

czullu:
Bring back ability to take UserAgent from config or environment when use new HttpClient Factory


jansenbe:
Good to indeed add that back, got lost with the work on unifying the http client bits

PR #206, opened 2021-03-05
fix if file is template in /libname/Forms folder and therefore has no ListItemId

czullu:
Upload of template files to the Forms folder of the Library did fail because of access to targetFile.ListItemAllFields.Id -
Added Check if have ListItemAllFields.


jansenbe:
Thx @czullu :-)

PR #228, opened 2021-03-18
Fix AudienceUriValidationFailedException exceptions when using AppCatalogScope.Tenant

jackpoz:
Fix AudienceUriValidationFailedException exceptions when using AppCatalogScope.Tenant scope but not being connected to the AppCatalog due to using the original Context instead of the AppCatalog cloned one.

The issue has been added in https://github.com/pnp/pnpframework/commit/7ccbab82ea10ba98ebb48b8899d2b6cce47addb4 most likely as a common copy&paste typo as "context" and "_context" are very similar.

This PR fixes for example calling
```powershell
Add-PnPApp -Path $Package -Overwrite -Scope Tenant -Connection $Connection
```
with a connection pointing to the Admin portal. Without this PR a AudienceUriValidationFailedException is thrown as the token used points to a different url than the AppCatalog.

### DIFF
--- a/src/lib/PnP.Framework/ALM/AppManager.cs
+++ b/src/lib/PnP.Framework/ALM/AppManager.cs
@@ -728,7 +728,7 @@ namespace PnP.Framework.ALM
 
             context.Web.EnsureProperty(w => w.Url);
 #pragma warning disable CA2000 // Dispose objects before losing scope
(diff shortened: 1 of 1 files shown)


jansenbe:
Good catch @jackpoz , thanks for submitting this fix 🥇

PR #230, opened 2021-03-22
Deserialze of Template XML does not load ClientSideComponentId on List-UserCustomAction

czullu:
The UserCustomAction on Lists could not be provisionied when needing ClientSideComponentId as it was not loaded by the Deserializer. As the Field did already exist in 2017, added the statement to both ListInstanceSerializer.cs (201705 / 201909)


jansenbe:
Nice catch @czullu , thanks for the fix.

PR #254, opened 2021-03-31
Avoiding get token attempt when connected using ACSAppOnly

KoenZomers:
Changes to avoid a NullRef exception on New-PnPSite when connecting using ACSAppOnly


jansenbe:
Closing as this will be fixed at PS level

PR #261, opened 2021-04-05
Added implementation of SPWebRequestExecutor that utilizes HttpClient

patrikhellgren:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | 

Added an implementation of WebRequestExecutorFactory and SPWebRequestExecutor that utilizes HttpClient instead of HttpWebRequest.  
This has the following benefits over the default implementation:  
- It makes it easier to prevent port exhaustion in for example larger scale provisioning scenarios
- It is faster
- You can take advantage of all the other benefits that comes with HttpClient

Usage example:
``` csharp
var authManager = new PnP.Framework.AuthenticationManager(clientId, certificate, tenantId);
var clientContext = authManager.GetContext(siteUrl);
clientContext.WebRequestExecutorFactory = new HttpClientWebRequestExecutorFactory(PnPHttpClient.Instance.GetHttpClient());
clientContext.Load(clientContext.Web, w => w.Title);
await clientContext.ExecuteQueryRetryAsync();
```

### DIFF
--- a/src/lib/PnP.Framework/Http/HttpClientWebRequestExecutorFactory.cs
+++ b/src/lib/PnP.Framework/Http/HttpClientWebRequestExecutorFactory.cs
@@ -1,11 +1,9 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
-using System.Text;
 using System.Threading.Tasks;
 using Microsoft.SharePoint.Client;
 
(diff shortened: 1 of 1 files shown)


jansenbe:
Nice work @patrikhellgren! Are you using this for all your CSOM contexts? Wondering if we should make this implementation the default one...

patrikhellgren:
@jansenbe I only made this implementation a couple of weeks ago but have been using it since then and my intention is to use it for all CSOM work hereafter since it seems to be running really well. I have tested it in our provisioning solution that runs in Azure Functions with some really large templates so I feel that I can say that at least the provisioning scenarios has been tested.  
If we were to make this the default implementation I still think that HttpClientWebRequestExecutorFactory should be public so that any HttpClient could be used if you had some special requirements, even though PnPHttpClient should be the default one.

PR #268, opened 2021-04-09
Minimal support for TEAMCHANNEL Template

czullu:
Adding the TEAMCHANNEL0 Template as XML and use RelatedGroupId as fallback for the {sitecollectionconnectedoffice365groupid} token during provisioning


jansenbe:
Nice improvement @czullu !

PR #274, opened 2021-04-13
add async version of GetACSAppOnlyContext

fowl2:
also improve it a bit to cache access tokens between `ClientContext`s

nb: note sure if all this `SharePointPnP.IdentityModel.Extensions.S2S.Protocols.OAuth2` stuff is supposed to be public?


jansenbe:
Hi @fowl2 , is there a particular reason why you want to keep this PR in draft?

fowl2:
@jansenbe I was hoping to get some time to finish that second commit, which I've now done. Hope it helps / any feedback appreciated.

src/lib/PnP.Framework/AuthenticationManager.cs
```diff
@@ -1061,13 +1064,33 @@ internal void ConfigureOnPremisesContext(string siteUrl, ClientContext clientCon
...
+            "Use AuthenticationManager.CreateACSAppOnlyAsync() and GetContextAsync()";
+
         /// <summary>
         /// Returns an app only ClientContext object
         /// </summary>
         /// <param name="siteUrl">Site for which the ClientContext object will be instantiated</param>
         /// <param name="appId">Application ID which is requesting the ClientContext object</param>
         /// <param name="appSecret">Application secret of the Application which is requesting the ClientContext object</param>
         /// <returns>ClientContext to be used by CSOM code</returns>
+        [Obsolete(GetACSAppOnlyContext_ObsoletionMessage)]
```
jansenbe, line 1093: @fowl2 : why are you marking the ACS methods as obsolete?

src/lib/PnP.Framework/AuthenticationManager.cs
```diff
@@ -867,7 +869,8 @@ public async Task<ClientContext> GetContextAsync(string siteUrl, CancellationTok
                     {
                         var context = GetAccessTokenContext(siteUrl, (site) =>
                         {
-                            return this.acsTokenGenerator.GetToken(new Uri(site));
+                            // HACK: sync over async
```
jansenbe, line 872: @fowl2 : what's the advantage here?

src/lib/PnP.Framework/Utilities/Context/ClientContextSettings.cs
```diff
@@ -10,20 +10,6 @@ internal class ClientContextSettings
         internal ClientContextType Type { get; set; }
         internal string SiteUrl { get; set; }
         internal AuthenticationManager AuthenticationManager { get; set; }
-
```
jansenbe, line 13: Why did you drop these?

jansenbe:
@fowl2 : this PR feels a bit like multiple changes are piled on, hard to verify all scenarios. 

@sebastianmattar : since you wrote the original ACS handling code I would appreciate your insights as well

src/lib/PnP.Framework/AuthenticationManager.cs
fowl2: because they're inconsistent with the other authentication methods, which complicates token caching

src/lib/PnP.Framework/Utilities/Context/ClientContextSettings.cs
fowl2: because they are not being used

src/lib/PnP.Framework/AuthenticationManager.cs
fowl2: it's incrementally converting the code to be async. This is equivalent to what the other auth methods are doing in `BuildClientContext`.  Perhaps that was just because MSAL has no sync apis(?) 

Presumably an more async-friendly alternative to `ExecutingWebRequest` will be added to the CSOM at some point and this can all be made async all the way through.  Maybe it'd be better to hold off on this part.

sebastianmattar:
> @fowl2 : this PR feels a bit like multiple changes are piled on, hard to verify all scenarios.
> 
> @sebastianmattar : since you wrote the original ACS handling code I would appreciate your insights as well

Only reason I left the ACS handling non-async was to make as few changes as possible to get my PR approved quickly. I think it is the right move to unify token handling.

jansenbe:
Hi @fowl2, Azure ACS is a legacy component and we're not planning to do anymore updates for it. Hope you can understand. Thanks.

fowl2:
1. Having a more similar API would help people migrate
2. The replacement auth methods don't have all the functionality of this one (hopefully) yet
3. As far as I know it hasn't officially been deprecated, probably due the #2

I understand there is still work involved in reviewing changes, but given the above I hope we can sort something out :)

fowl2:
I can do some more work to remove dead code if that helps.

PR #297, opened 2021-04-30
Fix conversion between PnPCore enum values and PnP.Framework enums

### LINKED ISSUE #266: [BUG] ClientSidePage Header always have TextAlignment set to "Center"
When I use `Get-PnPSiteTemplate` and include ClientSidePages the, the TextAlignment of the Header is always set to "Center".

The header text is set to align left in SharePoint, and the `Get-PnPClientSidePage` also return left

```powershell
(Get-PnPClientSidePage -Identity Column-Description).PageHeader.TextAlignment
# Left
```

jackpoz:
This PR fixes a conversion between PnPCore enum values and PnP.Framework enums. The old code does a C-style cast from the PnPCore enum to the PnP.Framework one but the 2 enums have the possible values defined in different order:

PnPCore:
```c#
  //
  // Summary:
  //     Alignment of the title in a page header
  public enum PageHeaderTitleAlignment
  {
      //
      // Summary:
      //     Page title is centered
      Center,
      //
      // Summary:
      //     Page title is left aligned
      Left
  }
```

PnP.Framework:
```c#
  public enum ClientSidePageHeaderTextAlignment
  {
      /// <summary>
      /// Align Left
      /// </summary>
      Left,
      /// <summary>
      /// Align Center
      /// </summary>
      Center,
  }
```

This means PageHeaderTitleAlignment.Left with value 1 gets converted to ClientSidePageHeaderTextAlignment of value 1, so ClientSidePageHeaderTextAlignment.Center .

The current code already tries to convert from an enum to another through .ToString() but parses it back to PnPCore enum instead of PnP.Framework.

To compare the old and the new code:
- old: PnPCore enum -> ToString() -> parse into PnPCore enum -> c-style cast to PnPFramework
- new: PnPCore enum -> ToString() -> parse into PnPFramework enum -> c-style cast to PnPFramework
While at it I updated 2 other similar fields `Type` and `LayoutType` that never showed any issue because the enums are defined in the same order in both projects.

In addition to this it might be worth to consider having the 2 enums with the same order in both projects.

Fix #266

### DIFF
--- a/src/lib/PnP.Framework/Provisioning/ObjectHandlers/Utilities/ClientSidePageContentsHelper.cs
+++ b/src/lib/PnP.Framework/Provisioning/ObjectHandlers/Utilities/ClientSidePageContentsHelper.cs
@@ -114,12 +114,12 @@ namespace PnP.Framework.Provisioning.ObjectHandlers.Utilities
                         var extractedHeader = new ClientSidePageHeader()
                         {
                             //Type = (ClientSidePageHeaderType)Enum.Parse(typeof(Pages.ClientSidePageHeaderType), pageToExtract.PageHeader.Type.ToString()),
-                            Type = (ClientSidePageHeaderType)Enum.Parse(typeof(PnPCore.PageHeaderType), (pageToExtract.PageHeader as PnPCore.PageHeader).Type.ToString()),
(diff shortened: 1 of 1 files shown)


jansenbe:
Good update @jackpoz , will also update the enum in PnP.Core

PR #299, opened 2021-05-01
Fix tokenize url in new document template

czullu:
make sure the url to the template document in NewDocumentTemplate-Json is tokenized in export.


jansenbe:
Good addition @czullu

Issue #292, opened 2021-04-28
[BUG] SSL connection cannot be established at random intervals


atomroflman:
When I am running a lot of provisioning templates (in my case ~400) I get the error: 

195 |  Invoke-PnPSiteTemplate -Path ".\[my Path] verfolgen  …                                                                                  |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                    | The SSL connection could not be established, see inner exception.                                                                                                                                                                                                                                                

That occours randomly every 2 to 40 executions.
If I am going to execute the same script again, same will happen with different templates. So rerunnig the failded ones will narrow down after some executions, but looking for each one that has falied is kind of annoying.

jansenbe:
@atomroflman : if you look at the inner exception can you then identify a common api call that throws this error or is it more at random?

patrikhellgren:
@jansenbe I have seen the same thing on a few occasions the last few weeks at least. This has occurred when running provisioning of large templates using just PnP Framework, so no PnP PowerShell and I always thought it was just a temporary thing but after reading this I did some testing.

### **tl;dr**
There seems to be some issues with how requests are handled server side when they probably should be throttled instead. The feeling is either you get the "SSL connection cannot be established" quite soon or you get "(400) Bad Request" later on. That is also the feeling I have from our production solutions where I have seen this. Question is if this is something that should be fixed server side or if this is something that must be handled in PnP Framework?

### **Detailed test examples**
I made a simple console app that just loops and reads the title from the connected web:

``` CSHARP
var authManager = new AuthenticationManager(clientId, certificate, tenantId);
var clientContext = await authManager.GetContextAsync(siteUrl);
for (int i = 0; i < 10000; i++)
{
  clientContext.Load(clientContext.Web, w => w.Title);
  await clientContext.ExecuteQueryRetryAsync();
  Console.WriteLine($"{i} - {clientContext.Web.Title}");
}
```
Every time I ran this it got to request number 1100 - 1400 and then the exception "SSL connection cannot be established" was cast. The inner exception is "An existing connection was forcibly closed by the remote host."
I have tried it with different tenants with the same result, each and every time I ran it I got the exception somewhere between request 1100 and 1400. Sometimes even sooner.

I have also tested with just using the default ExecuteQueryAsync with the same result.

[...]

PR #301, opened 2021-05-01
Explicit Catch of System.IO.IOException in ExecuteQueryImplementation…

czullu:
#292 Took the Idea with TraceCorrelationId and write this to log by handling IOExpection explicit,
also handle TimeOut with explicit message. Lets see if i can get some input for SSL from Application Insight like this.


jansenbe:
Thanks @czullu , adding more logging and resilience when it comes to socket exceptions/ timeouts makes sense.

PR #310, opened 2021-05-06
Fix an issue with possible access of null serverobject in FieldUtilities.FixLookupField

### LINKED ISSUE #309: FieldUtilities.FixLookupField throws exception when Lookup's "List" attribute is not set
I encountered an issue when applying a PnP template, since I had a lookup field with no _List_ attribute set.
This causes `FieldUtilities.FixLookupField` to throw an exception, since `GetListByUrl` returns an object which has `ServerObjectIsNull == true`.

https://github.com/pnp/pnpframework/blob/1f3e143c26d3e093d6a632708eb39ff265dfb1f9/src/lib/PnP.Framework/Provisioning/ObjectHandlers/Utilities/FieldUtilities.cs#L18-L23

@jansenbe I'm more than willing to send a PR, but was wondering what would be the best place to fix this.

Should I update `FieldUtilities.FixLookupField` to check for `ServerObjectIsNull` (which would be the least breaking change I assume), or should `ListExtensions.GetListByUrl` return `null` when `ServerObjectIsNull` is `true`? (which is probably not ideal in order to not break anything down the line?)

NickSevens:
Fixes #309.

Alternative would be to change `ListExtensions.GetListByUrl` to return `null` when `ServerObjectIsNull == true`, as I assume this issue could occur on other places as well... That would however be more of an impact than wanted I assume. :)

Not sure what's the best approach?

### DIFF
--- a/src/lib/PnP.Framework/Provisioning/ObjectHandlers/Utilities/FieldUtilities.cs
+++ b/src/lib/PnP.Framework/Provisioning/ObjectHandlers/Utilities/FieldUtilities.cs
@@ -16,7 +16,7 @@ namespace PnP.Framework.Provisioning.ObjectHandlers.Utilities
                 if (!Guid.TryParse(listAttr, out Guid g))
                 {
                     var targetList = web.GetListByUrl($"/{listAttr}");
-                    if (targetList != null)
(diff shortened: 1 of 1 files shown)


NickSevens:
@jansenbe any feedback you can give on this?

PR #318, opened 2021-05-10
Fixed a bug that updateChildren value was ignored in AddContentTypeToList

antonsmislevics:


jansenbe:
Hi @antonsmislevics : great to see you again, that has been a while. Hope your doing good.

PR #326, opened 2021-05-15
Feature: improve alias naming for modern sites.

gautamdsheth:
Improve alias naming for modern sites by removing special and accented characters.


jansenbe:
Thanks @gautamdsheth , good stuff!

PR #345, opened 2021-05-24
Fix for valid characters and lower case in site alias

patrikhellgren:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Fix for an issue that I noticed after a recent PR that changed the behavior of site alias naming. The characters - and _ are valid in site alias and it doesn't have to be lowercase. When creating a new team site using the SharePoint Online start page you are allowed to use - and _ and it doesn't convert site alias to lowercase so I think we should follow the same rules as that oob wizard. We have used the character - in site aliases for a long time and that broke after the recent PR merge.


jansenbe:
@patrikhellgren , thanks for pointing this out. I'll have a look

PR #382, opened 2021-06-16
Fix so that teams apps are added before channels

patrikhellgren:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| New sample?      | no
| Related issues?  | 

Fix for an issue that appears when trying to provision a teams tab with a custom app from the app catalog. The app needs to have been added to the team before a tab that references the app can be added.


jansenbe:
Nice catch @patrikhellgren !

PR #384, opened 2021-06-17
Added new provisioning token for using sequence site tokens in global template context

patrikhellgren:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| New sample?      | no
| Related issues?  | 

Added a new token called sequencesitetoken to be used to get to specific sequence site tokens in the global template context. This way you can for example use site tokens in Teams channel messages or Teams tabs configurations which was not possible before.

Usage example: ```{sequencesitetoken:MYID:listid:My List}```

I would also like to contribute documentation for this but cannot find anything around the tokens except this one for PnP-SItes-Core which has been archived: [Office 365 Developer PnP Core Component Provisioning Engine Tokens](https://github.com/pnp/PnP-Sites-Core/blob/master/Core/ProvisioningEngineTokens.md). Where should the documentation around the tokens be kept?


jansenbe:
@patrikhellgren : thanks for this change. Regarding docs: today there's no pnp framework documentation...not ideal, but that's due to the resource constraints. For PnP Core SDK we do have decent docs and once (not yet planned) we port the provisioning engine we'll write detailed provisioning engine docs. In the meanwhile I would recommend writing the docs as code comments in the SequenceSiteTokenToken class

PR #391, opened 2021-06-25
Issue 390 fixed in CreateOrUpdateTeamFromGroupInternal. 

luismanez:
Now the existing retry code will work as expected. When the CreateOrUpdateTeamFromGroupInternal calls the GraphHelper

```cs
teamId = GraphHelper.CreateOrUpdateGraphObject(.....)
```

If the returned teamId is null or empty, the existing retry code will be hit. The _wait_ variable is set to false only in the teamId is not null

```cs
if (!string.IsNullOrEmpty(teamId))
{
    wait = false;
}
```


jansenbe:
Good catch @luismanez , thanks for sharing the fix!

Issue #372, opened 2021-06-09
Save ProvisioningHierarchy in 1.5 creates some times wrong Xml-Element Order


czullu:
the strange issue is, that saving the ProvisioningHierarchy with XMLOpenXMLTemplateProvider to a pnp-file creates sometimes the XML-Elements in wrong order which make the Template fail when load again in XmlPnPSchemaBaseSerializer.GetValidationResults

What did happen multiple time is that pnp:Templates is first followed by pnp:Preferences, pnp:Localization and the pnp:Teams

@PaoloPia i went back to commit e40396c4b3ef86d974ed39f0fd6426599a652cda - before schema changes. With that Version i can not reproduce the issue.

jansenbe:
@czullu : I've seen that before and so far, we've not been able to reproduce. Our testing a while ago indicated that it only failed in .NET Core...are you using that?

czullu:
Yes, It happens in .net core 3.1 for me and I was also not really able to track it down as it seems like 1 out of 10 having this issue.
I fixed it on my side by reorder the elements on import. Let me provide this and you can decide if that's good enough for now.

Holen Sie sich Outlook für Android<https://aka.ms/AAb9ysg>

________________________________

andikrueger:
I'm facing this issue too. I see XML files with wrong sequences too>

```xml
  </pnp:Templates>
  <pnp:Preferences Generator="PnP.Framework, Version=1.6.0.0, Culture=neutral, PublicKeyToken=null" />
</pnp:Provisioning>
```

The xml above was created in a PowerShell 7.1 session using PnP.PowerShell to export a list as a template. The xml file is generated wrong in all my test runs.

Adding addtional data to the templates fails during schema validation with the following error: 
```powershell
Add-PnPDataRowsToSiteTemplate: 
Line |
  36 |      Add-PnPDataRowsToSiteTemplate -Path $tempFile -List 'PnPTestList' …
     |      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | The element 'Provisioning' in namespace 'http://schemas.dev.office.com/PnP/2021/03/ProvisioningSchema'
     | has invalid child element 'Preferences' in namespace
     | 'http://schemas.dev.office.com/PnP/2021/03/ProvisioningSchema'. List of possible elements expected:
[...]

PR #398, opened 2021-07-04
make sure the XML-Elements are in the right order for schema validation

czullu:
This fix is a suggestion on we possible handle #372. It has the benefit, that we can use generated templates even when the order of the XML-elements is wrong.

### DIFF
--- a/src/lib/PnP.Framework/Provisioning/Providers/Xml/XmlPnPSchemaBaseSerializer.cs
+++ b/src/lib/PnP.Framework/Provisioning/Providers/Xml/XmlPnPSchemaBaseSerializer.cs
@@ -68,6 +68,27 @@ namespace PnP.Framework.Provisioning.Providers.Xml
             // Load the template into an XDocument
             XDocument xml = XDocument.Load(template);
 
+            List<XElement> orderedElements = new List<XElement>();
+            var Preferences = xml.Root.Elements().FirstOrDefault(e => e.Name.LocalName == "Preferences");
+            if (Preferences != null) orderedElements.Add(Preferences);
+            var Localizations = xml.Root.Elements().FirstOrDefault(e => e.Name.LocalName == "Localizations");
+            if (Localizations != null) orderedElements.Add(Localizations);
+            var Tenant = xml.Root.Elements().FirstOrDefault(e => e.Name.LocalName == "Tenant");
+            if (Tenant != null) orderedElements.Add(Tenant);
+            var Templates = xml.Root.Elements().FirstOrDefault(e => e.Name.LocalName == "Templates");
+            if (Templates != null) orderedElements.Add(Templates);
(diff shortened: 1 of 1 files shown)


jackpoz:
Just out of curiousity, is XML itself so strict that it requires elements order to be the exact same as the schema or is it the .NET Core implementation that has some issues (or even the pnpframework code itself) ?

czullu:
> Just out of curiousity, is XML itself so strict that it requires elements order to be the exact same as the schema or is it the .NET Core implementation that has some issues (or even the pnpframework code itself) ?

it depends on the schema definition - the way it's defined in the XSD-Files in 
\pnpframework\src\lib\PnP.Framework\Provisioning\Providers\Xml\ProvisioningSchema-.....xsd
it has to be in order.

andikrueger:
@czullu Is it really a solution to reorder the elements? The xml would still not be valid in regards of the schema definition. 

I see this error very persistent within my dev environment. I reordered the element manually and ran additional commands to modify the template. After the command completed, the order was gone and the elements back in back order. I will investigate a bit further.

czullu:
@andikrueger The fix does work for me but it's not part of the solution. The fix would reorder the elements on the secound level of the xml according to here https://github.com/pnp/PnP-Provisioning-Schema/blob/master/Samples/ProvisioningSchema-2021-03-FullSample-01.xml after reading the XML to Memory but before XML-Schema-Validation. As the XML-Schema validation check for each and any element and it's attributes you need to make sure all is in place and order - in your case manually as i understand. For me since last updates of dotNet (3.1/5.0) the missorder happen when writing the XML to the filestream in pnp.framework happens less often - so you might try updating to KB5004331/KB5006191.

andikrueger:
I was missing the two updates on my dev system and will do some more testing now...

I did some more digging through the .cs classes that were generated from the .xsd files. The classes do look fine. What could be considered to enforce the sequence is the `order` attribute. This could also be generated automatically by using `xsd.exe` with some more options. https://docs.microsoft.com/en-us/dotnet/standard/serialization/xml-schema-definition-tool-xsd-exe#xsd-file-options
Maybe this could resolve the initial issue.

andikrueger:
Short Update: KB5006191 did change the behavior and I was able to successfully extract the template. The XML structure looks fine now.

```xml
<pnp:Provisioning xmlns:pnp="http://schemas.dev.office.com/PnP/2021/03/ProvisioningSchema">
<pnp:Preferences Generator="PnP.Framework, Version=1.6.0.0, Culture=neutral, PublicKeyToken=null"/>
<pnp:Templates ID="CONTAINER-TEMPLATE-08DFCAC23287403A97A9663B0A0BB174">
```

jansenbe:
Closing this PR as the only right fix is ensuring you've .NET patched. Thanks for the research on this @czullu  and @andikrueger !

PR #470, opened 2021-09-23
CreateFolderInList Handle Error "To update this folder, go to the channel in Microsoft Teams"

czullu:
creating subfolder to channel folder does fail as ContentType, Properties and PropertyBag Update on Channel Folder is not allowed.
The suggested fix does catch the specific error, logs it as warning and continues to create folder structure


czullu:
@jansenbe  Hi Bert, since you didn't merge this pull request, I suspect you have some concerns about handling the update error for Team Channel Folders. The only other way I can think of is to pre-check if the "vti_x005f_teamchannelurl" property is present in the folder, and skip any possible updates from the template for the ChannelFolder itself to create the subfolders as specified in the template. I can change the code if this is the way you would prefer it.

jansenbe:
@czullu : this one would need some more review time, hence I did not yet merge it. Hope to have more time in the coming days

PR #478, opened 2021-10-03
Adding retry logic on throttling to RESTUtilities.ExecuteGetAsync

KoenZomers:
Currently no retry logic is implemented when throttling would occur. Adding this logic. Reported through https://github.com/pnp/powershell/issues/949


jansenbe:
Hi @KoenZomers ,

We do have retry logic implemented at the httpclient level, see https://github.com/pnp/pnpframework/blob/dev/src/lib/PnP.Framework/Http/RetryHandler.cs for the implementation being used. Can we reproduce issues with this generic retry handling?

PR #479, opened 2021-10-03
Added additional check to avoid exception when footer not enabled on a site

KoenZomers:
When currently trying to get the footer Title or Image providing a site where the footer is disabled, it will throw a key not found in dictionary exception. This stops `Get-PnPFooter` from working. This PR will fix that issue.


jansenbe:
Nice catch @KoenZomers !

Issue #500, opened 2021-10-15
EnsureFolderPathAsync with # or % fails with File not Found


jimmywim:
Looks like the folder creation part of `EnsureFolderPath` needs to be tweaked to use ResourcePaths, using `ListItemCreationInformationUsingPath` instead of `ListItemCreationInformation`.

I've done this with a locally copied out version of the method, seems to work nicely.

I'll submit a PR for review.

PR #501, opened 2021-10-15
Issue fix 500, unable to ensure folder with path characters

jimmywim:
This fixes an issue where you use `EnsureFolderPath/Async` to ensure a folder path that contains path characters, `%` and `#`.


jimmywim:
Fixes issue: https://github.com/pnp/pnpframework/issues/500

PR #531, opened 2021-11-24
Fixing issue when trying to export datarow for list instances

magarma:
There is an issue in the code that is preventing to export properties for files from libraries when there are several content types.
Not sure if affects to more cases. 
In line 587, code is adding ContentTypeId to the pnpFile.Properties. In line 613, was trying to add again the same property key to the dictionary, that was raising an issue.
Changing the way to add properties to the dictionary, it will overwrite it if exists.
image


jansenbe:
Good change @magarma , thanks for helping make PnP Provisioning better 💗

Issue #444, opened 2021-08-12
"TeamsAppId" Property Missing from Extracted Schema File for Teams.


BaronSparky:
### Reporting an Issue or Missing Feature
Issue

### Expected behavior 
To be able to extract a Teams configuration from an existing Team and then re-provision from that extracted template.

### Actual behavior 
Trying to re-provision from an extracted template results in errors:
```
Invoke-PnPTenantTemplate : The required attribute 'TeamsAppId' is missing.
At line:1 char:1
+ Invoke-PnPTenantTemplate -Path .\sometemplate.xml
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : SyntaxError: (:) [Invoke-PnPTenantTemplate], XmlSchemaValidationException
    + FullyQualifiedErrorId : TEMPLATENOTVALID,PnP.PowerShell.Commands.Provisioning.Tenant.InvokeTenantTemplate
```
This relates to the "TeamAppId" property missing from the extracted schema for each of the "Tabs" defined. (https://docs.microsoft.com/en-us/graph/teams-configuring-builtin-tabs)
```
<pnp:Tab DisplayName="WebSite" ID="fdf0fc44-4eca-43e8-b8e1-932f7a60f1f7" Remove="false">
  <pnp:Configuration EntityId="" ContentUrl="https://www.somewebsite.co.uk" RemoveUrl="" WebsiteUrl="https://www.somewebsite.co.uk" />
</pnp:Tab>
```
needs to be similar to:
```
<pnp:Tab DisplayName="WebSite" TeamsAppId="com.microsoft.teamspace.tab.web" ID="fdf0fc44-4eca-43e8-b8e1-932f7a60f1f7" Remove="false">
  <pnp:Configuration EntityId="" ContentUrl="https://www.somewebsite.co.uk" RemoveUrl="" WebsiteUrl="https://www.somewebsite.co.uk" />
</pnp:Tab>
```
### Steps to reproduce behavior
Extract a PnP Tenant Template using a configuration file that looks like:
```
{
    "$schema": "https://aka.ms/sppnp-extract-configuration-schema",    
    "tenant": {
        "teams": {          
            "includeGroupId": false,            
            "includeMessages": false
        }
    },
    "handlers": [
        "Tenant"
    ]
}
```
Issue the command for the SPO site that supports the Team and save it as an XML schema file::
```
[...]

PR #532, opened 2021-11-26
Fix issues exporting channels and tabs with TeamsAppId and Description

magarma:
Hi,
When exporting Team Channels and Tabs, TeamsAppId for the Tab is never retrieved, also, when the Channel Description is null, the Channel description attribute is missing.
Both attributes are required and if they are missing, trying to apply the template, it fails.
image

There is an opened issue about it.
#444 

I've changed the way that the solution gets the app id to reuse an exiting method that was already in the class. 
This is related with this change
https://github.com/pnp/pnpframework/pull/189

Now, both attributes are provided.
image


jansenbe:
Nice work @magarma !

sympmarc:
Until this makes it into the released version, is there a trick we can use to get the correct `TeamsAppId` to add manually?

magarma:
There is a nightly build, you could try to use that until the 1.8 is released.

Issue #521, opened 2021-11-11
Sensitivity Label not applied when using TeamSiteCollectionCreationInformation object to Create Team Site


ashu57:
Hello , 

I have validated that this method doesn't support yet applying the sensitivity label to the SharePoint Team Site/Group.

Class
PnP.Framework.Sites.SiteCollection CreateAsync

Method:
public static Task<ClientContext> CreateAsync(ClientContext clientContext, TeamSiteCollectionCreationInformation siteCollectionCreationInformation, int delayAfterCreation = 0, int maxRetryCount = 12, int retryDelay = 10000, bool noWait = false, string graphAccessToken = null, AzureEnvironment azureEnvironment = AzureEnvironment.Production);

Please check and include this & also advise if any alternative exist in PnP Framework to apply sensitity label for team site.

NikoMix:
Hello Ashu,

thanks for your Feedback. Please note, that I'm not representing Microsoft on this Issue, while working for Microsoft.

The Parameter of Type TeamSiteCollectionCreationInformation should have a SensitivityLabel and SensitivityLabelId property.
They are however not available through the constructor.

Do you mean, that you created a Site through CreateAsync, providing a value to this property but it did not apply this correctly?

ashu57:
Hello @NikoMix  ,
Indeed , if i set classification Label property of the object TeamSiteCollectionCreationInformation then the Sensitivity Label is not applied to the site after creation.
In addition if i check the source code of the CreateAsync method with signature mentioned above, it does not have implementation for applying sensitivity label .

thanks for checking.

NikoMix:
@ashu57  thank you for clarification.

So I've checked again and indeed, it seams like there is no SensitivityLabel set. I'll try to propose a PR for it.
Just to be sure, this will address your Issue - what would be your expectation based on the set SensitivityLabel/SensitivityLabelId? As the TeamSite is creating a modern SharePoint Site with a Unified Group and can create a Teams Team based on that Group.

The Label should be applied to:

- [ ] Modern Site
- [ ] Unified Group
- [ ] Team

Personally I'd would like to have three parameters, to be able to set different labels based on the resource - but not sure, if they will even be synced anyway.

jansenbe:
Fix has been merged, thx @NikoMix ! 

@ashu57 :  please try with the next nightly release and see if your issues are gone. If so please close this issue

renabar:
@NikoMix  @jansenbe 
[...]

PR #536, opened 2021-11-29
Added Sensitivity Labels to Unified Group Creation

NikoMix:
Proposed fix for #521


jansenbe:
Nice change @NikoMix , thanks for helping the community with this fix!

magarma:
We're are not able to create Groups with app-only tokens.
This is the request is doing even we are not apply labels.
image

Could we revert this? or maybe it would help checking assignedLabels is not empty before sending the request and don't provide the parameter.

PR #544, opened 2021-12-09
521

NikoMix:
This is to address the issue with no labels applied #521 

Group could be created through app-only context again. Also added XML comment, that labels will break app-only context.


magarma:
Hi @NikoMix, 
I've given a try to this change in my dev environment and I think that there could be a mistake...
As you can see, you are checking if assignedLabels is null or empty, but I think that you should check the variable labels, 
asignedLabels is coming with Guid.Empty. Look at my screenshot.
image

NikoMix:
@magarma did you specify a label to the array? Without it gave a null reference exception, but this should be fixed now - the default for it is null

```csharp
Guid[] assignedLabels = null
```

magarma:
@NikoMix but in your previous PR your did something like this...
image

As you can see, the Guid.Empty is being sent as parameter.

NikoMix:
I've tested based on the method itself

```csharp
    var group = UnifiedGroupsUtility.CreateUnifiedGroup("test", "this is description", "test2", 
        token.AccessToken, assignedLabels: new [] { Guid.Empty });
```

let me setup another test for the CreateTeamSiteViaGraphAsync, I'll post update here once finished.

magarma:
Tested again with your update and it works ok now for me.

PR #557, opened 2021-12-23
Fix to trim unaccepted chars from SiteAlias

gautamdsheth:
Fix to trim unaccepted chars from the SiteAlias as well.

We already fixed it for Alias property, added the same fix for SiteAlias property


gautamdsheth:
Closing it , reopening it from a new feature branch :)

PR #562, opened 2022-01-03
Additional changes for fix #498

### LINKED ISSUE #498: Missing HeaderLayoutType
Hello,

There are two out out of four HeaderLayoutTypes being handled in those switch statements:

https://github.com/pnp/pnpframework/blob/ac4dc39f564fcaa55e140bd7d778149360a265f7/src/lib/PnP.Framework/Provisioning/ObjectHandlers/ObjectSiteHeaderSettings.cs#L27

https://github.com/pnp/pnpframework/blob/ac4dc39f564fcaa55e140bd7d778149360a265f7/src/lib/PnP.Framework/Provisioning/ObjectHandlers/ObjectSiteHeaderSettings.cs#L60

"Extended" and "Minimal" (already present in HeaderLayoutTypes enum) are missing from those two switches

Could you add those?

Thanks in advance!

fzbm:
The previously made changes for fix #498 were missing changes required for provisioning the header information.

Additional changes:

* Removed case for `HeaderLayoutType.Standard` in `ExtractObjects`,  because `default` already handles this.
* Removed `scope` variable, because it is not used.
* Removed the `Emphasis` type reference from `Enum.TryParse`, because  the `out` variable already specifies the type.
* `ProvisionObjects` now exits as soon as possible if no changes are required.
  * This also moves a possible server call for loading the web URL to a state, were this information would be required.
* Introduced some (single) blank lines to increase the readability of the code.

### DIFF
--- a/src/lib/PnP.Framework/Provisioning/ObjectHandlers/ObjectSiteHeaderSettings.cs
+++ b/src/lib/PnP.Framework/Provisioning/ObjectHandlers/ObjectSiteHeaderSettings.cs
@@ -73,7 +73,7 @@ namespace PnP.Framework.Provisioning.ObjectHandlers
                     return parser;
                 }
 
-                web.EnsureProperties(w => w.Url);
+                web.EnsureProperties(w => w.HeaderLayout);
 
                 switch (template.Header.Layout)
                 {
--- a/src/lib/PnP.Framework/Provisioning/ObjectHandlers/ObjectSiteHeaderSettings.cs
+++ b/src/lib/PnP.Framework/Provisioning/ObjectHandlers/ObjectSiteHeaderSettings.cs
@@ -17,13 +17,15 @@ namespace PnP.Framework.Provisioning.ObjectHandlers
 
         public override ProvisioningTemplate ExtractObjects(Web web, ProvisioningTemplate template, ProvisioningTemplateCreationInformation creationInfo)
         {
-            using (var scope = new PnPMonitoredScope(this.Name))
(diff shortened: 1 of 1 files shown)


PedroMordeP:
Hi was looking at this pull request and at the line 76 on ProvisionObjects
`web.EnsureProperties(w => w.Url); `
should be changed to                
`web.EnsureProperties(w => w.HeaderEmphasis, w => w.HeaderLayout, w => w.MegaMenuEnabled);`

??

PedroMordeP:
If we don't do the `web.EnsureProperties(w => w.HeaderEmphasis, w => w.HeaderLayout, w => w.MegaMenuEnabled);`

I was getting this issue
**_The property or field 'HeaderLayout' has not been initialized. It has not been requested or the request has not been executed. It may need to be explicity request._**

and the w.Url is not used :)

fzbm:
@PedroMordeP You are right. I also stumbled up on this exception during previous tests. A temporary fix is to load the web or specifically the `HeaderLayout` property before applying the template. I changed this and create an additional commit. A test of the change worked just fine. I applied a template to a freshly created `ClientContext` and PnP changed the header layout type as expected.

Also loading `HeaderEmphasis` and `MegaMenuEnabled` is not required, because `HeaderLayout` is the only property which might not get set. The two other will definitely be set before accessing the value.

`ExecutePostAsync` requires the web URL. While the method ensures the property itself, loading it beforehand will save us an additional round trip to the server.

PR #567, opened 2022-01-07
Fix site footer

czullu:
this change overlaps with open pull request #498 / #562 but includes the handling for 
ShowSiteTitle = !web.HideTitleInHeader
as well. i did try to align for easier merging


jansenbe:
@czullu : I'm not yet merging this one. The key reason is (and it's not just due to your changes) is that actually we should use the SetChromeOptions over updating the respective web properties as updating these web properties requires full control on the site and therefore might fail. I've re-implemented branding in PnP Core SDK (see https://pnp.github.io/pnpcore/using-the-sdk/branding-intro.html and https://github.com/pnp/pnpcore/tree/dev/src/sdk/PnP.Core/Model/SharePoint/Branding) without updating the web. This approach is aligned to what is offered by the UI and can maybe serve as inspiration. Possibly some of the PnP Core SDK logic can be reused/extended for this purpose.

Also adding @fzbm and @PedroMordeP  as they've been active on site header/footer configuration

czullu:
Ok - was not aware that there is a difference but can change to use pnpcorecontext as in ObjectClientSidePages. 
i can do the change next weekend and let you know.

eduardpaul:
Hi, will this PR fix this https://github.com/pnp/pnpframework/issues/646 ? Sorry opened the issue without checking active PR's.

jansenbe:
closing as too old

PR #570, opened 2022-01-11
Added missing values for Site Header Layout on Schema 2020.02

PedroMordeP:
Fix error when using template schema 2020.02 and changing the Header layout to "Minimal" or "Extended" gives an error.

_Template schema does not exist | Template is not valid
The 'Layout' attribute is invalid - The value 'Minimal' is invalid according to its datatype 'String' - The Enumeration constraint failed._


jansenbe:
Hi @PedroMordeP , these values are part of our next schema version (ProvisioningSchema-2021-03), why do you want to back port them? We recommend upgrading to the latest schema instead of updating older schemas. 

FYI: @PaoloPia

PedroMordeP:
> Hi @PedroMordeP , these values are part of our next schema version (ProvisioningSchema-2021-03), why do you want to back port them? We recommend upgrading to the latest schema instead of updating older schemas.
> 
> FYI: @PaoloPia

Hi @jansenbe, we are still using this previous schema using the [PnP-Powershell](https://github.com/pnp/PnP-PowerShell/) versions but we are planning to upgrade our product to use the latest framework.

Meanwhile, we need to use the 'Minimal' value on the Layout, so I saw this value was not available on this schema and thought to add them since are only 2 values and hopping this to be accepted.

We are still using a older version because of an AccessToken issue of only working on specific older versions. I run on PowerShell a test yesterday to debug this and tested with the new framework version 1.9.0 and is working, now it gets the context using the connect-pnponline with the AccessToken.
So it is planned to upgrade to this version and run more tests on our side.

PedroMordeP:
We can discard this pull request. I end up updating our engine to the newer schema but I'm waiting for the release to fix the Header Layout issue https://github.com/pnp/pnpframework/pull/562.
@jansenbe Can you have a date for the new release?

PR #603, opened 2022-02-14
Fix copy/update of taxonomy fields on modernization - not all fields …

erosu-hab:
There is currently an issue where only one of the taxonomy fields is copied to the target page, even though more than one field is specied in MetaData element of the PageLayout.
For example, we have two taxonomy fields NewsCategory and NewsRegionMulti which they should be copied in their corresponding target fields:

<MetaData>
	<Field Name="Title" TargetFieldName="" Functions="" />
	....
	<Field Name="NewsCategory" TargetFieldName="NewsCategory" Functions="" />
	<Field Name="NewsRegionMulti" TargetFieldName="NewsLocation" Functions="" />
	....
</MetaData>

In this scenario, only NewsLocation gets saved to the target. NewsCategory does not get saved.

From my tests, the mapping between source and target fields works as expected, and the values to be saved in the target fields are calculated properly. However, the calls in lines 648-650 in PublishingMetadataTransformator.cs to persist the taxonomy fields only save the last field, which in our case is NewsLocation.

I think the issue is related to the context beign used in GetAllTermsFromTermSet (TermTransformator.cs) to retrieve the term sets for cache. If instead of using the current context we use a cloned context, it appears that all taxonomy fields are beign saved correctly.

### DIFF
--- a/src/lib/PnP.Framework/Modernization/Transform/TermTransformator.cs
+++ b/src/lib/PnP.Framework/Modernization/Transform/TermTransformator.cs
@@ -263,31 +263,34 @@ namespace PnP.Framework.Modernization.Transform
 
             try
             {
-                TaxonomySession session = TaxonomySession.GetTaxonomySession(context);
-                TermStore termStore = session.GetDefaultSiteCollectionTermStore();
-                var termSet = termStore.GetTermSet(termSetId);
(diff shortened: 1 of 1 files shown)


jansenbe:
Using a new context will avoid possible internal caching, merging this one

PR #607, opened 2022-02-17
Retrieve the target list by url for handler ObjectListInstanceDataRows

### LINKED ISSUE #606: Object Handler ListInstanceDataRows does not work for some native lists
I get a site template from an **English** Communication Site with the handler **List** and the parameter **ListsToExtract** = **Events**.
Then, I add to the site template the data rows from this list.

When applying this site template to my site in **French**, it returns an error on applying the data rows because "the list Events does not exist".
This is because the native list **Events** in French is called **Événements**.
The same problem exists for some native lists/libraries like Style Library (Bibliothèque de styles), Site Assets (Pièces jointes), Site Pages (Pages du site),  etc...

The source of the issue is in `ObjectListInstanceDataRows.cs`: the target list is retrieved with `GetByTitle` instead of retrieve it with `GetListByUrl`.

I will do a Pull Request to fix this behavior.

BollietMZK:
The target list was retrieved with `GetByTitle` instead of retrieve it with `GetListByUrl`.
This behavior causes errors for native lists (like **Events**) whose title is translated according to the language of the site (**Événements** in French).

Fix #606


jansenbe:
Good fix @BollietMZK , thanks for sharing this one.

PR #615, opened 2022-03-03
Added user not found warning

robi26:
Added a more specific warning when a user could not be resolved.


jansenbe:
@robi26 : I'm not merging this one as the chosen fix is not completely implemented. To real solution should be to have a "fallback" account for in case a source account does not exist anymore.

PR #628, opened 2022-03-20
Fix exception thrown when a Taxonomy Term owner was not found

jackpoz:
Fix an exception happening when exporting a PnP template with `Get-PnPSiteTemplate -IncludeSiteCollectionTermGroup -Handlers TermGroups` and applying it to another tenant. The Owner field of `<pnp:Term Owner="...">` would point to the owner of the source tenant, which cannot be found on the destination tenant.

This PR adds a "user exists" check and logs a warning when the user is not found.

This is the error happening without this PR:
image

I couldn't find a generic "DoesUserExist()" method in the library, please let me know if I missed it and if I should update the PR


jansenbe:
Good fix, thanks for sharing @jackpoz !

PR #636, opened 2022-04-06
Implement TermGroup UpdateBehavior.

rompenar:
Update TermGroup properties: Description, Contributors, Managers.
Update TermSet properties: Name (same id), Description, IsOpenForTermCreation, IsAvailableForTagging, CustomProperties.
Update Term properties: Name (same id), Description, IsAvailableForTagging, Labels, CustomProperties, LocalCustomProperties.

### DIFF
--- a/src/lib/PnP.Framework/Provisioning/ObjectHandlers/Utilities/TermGroupHelper.cs
+++ b/src/lib/PnP.Framework/Provisioning/ObjectHandlers/Utilities/TermGroupHelper.cs
@@ -14,83 +14,39 @@ namespace PnP.Framework.Provisioning.ObjectHandlers.Utilities
         {
             List<ReusedTerm> reusedTerms = new List<ReusedTerm>();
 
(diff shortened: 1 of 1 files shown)


pschaeflein:
@rompenar - I'm struggling to understand the change. You've created a method named CreateTermGroup, and this method returns a tuple that includes a boolean. And the boolean indicates whether the group is new? This seems odd - I would expect a method with that name to create a term group, or throw if something goes wrong.

Was there an error you were fixing? Or is this an optimization of the code? Some more details would be helpful.

Thanks.

rompenar:
@pschaeflein

The boolean return value is needed for the "modelTermGroup.Name == "Site Collection"" condition in the CreateTermGroup. When the group to create is the site collection one (<pnp:TermGroup Name="{sitecollectiontermgroupname}" ...), we use the methods termStore.GetSiteCollectionGroup which does not set the Contributors and Managers properties. We use the boolean value to avoid update when it's a new group, but to force it for new site collection group.

This change was made to implement the UpdateBehavior property on TermGroup, TermSet and Term object.

PR #667, opened 2022-05-27
Fixed bug with xml serializer, team drive and team photo

danielpastoor:
Info:
I'm using pnp.framework for provisioning teams with "Application" permissions.

But when I do that then i encounter some issues:
1:
If you use application premissions for extracting the teams photo then you will get an error from the Graph Api as the background call in the Graph Api has a bug that it can't be Accessed by "Application" permissions.
Solution:
Add a back-up call to the group photo instead of the teams photo if the teams photo call fails.

2:
There is a bug in the template provisioning for hierarchy templates. When there is no "ProvisioningTemplate" then it still added a "pnp:Templates". And this will give a invalid xml error
Solution:
Check if there are any "ProvisioningTemplate"

3:
Initialize of the drive for the general channel. I got an error everytime that it isn't available when I try to provision.
Solution:
Add a try and catch so it can be swallowed and don't stops the provisioning


danielpastoor:
Pull request now also include a fix for Issue: 
https://github.com/pnp/pnpframework/issues/671

jansenbe:
Hey @danielpastoor , can you please split this PR into 3 PR's. When you combine different fixes in a single PR it gets harder to evaluate and merge these, a best practice is one PR per change. Also @patrikhellgren did a fixes to the Teams provisioning which might solve issue 3 for you...please do a get latest and verify if issue 3 is still a problem.

danielpastoor:
> Hey @danielpastoor , can you please split this PR into 3 PR's. When you combine different fixes in a single PR it gets harder to evaluate and merge these, a best practice is one PR per change. Also @patrikhellgren did a fixes to the Teams provisioning which might solve issue 3 for you...please do a get latest and verify if issue 3 is still a problem.

Hi Jansenbe,

I understand the problem I Will convert it to 3 PR's and check if the problems are still there

PR #695, opened 2022-07-18
Converted Hierachy provisioning and extraction to async

danielpastoor:
I have converted the hierachy handlers to async and added a overload for ApplyTenantTemplate so that it is still posible to call it not async.

The reason why I did this is because if had problems with the creation of the graph client.

And also I'm not a big fan of the .GetAwaiter() function as it fakes the await and can give some problemes.


pschaeflein:
@danielpastoor - Can you please elaborate on the "problems with the creation of the graph client"? I don't follow how creating the client depends on the synchronicity of the method.

danielpastoor:
It is not relevant any more. it is working again. I will close the pull-request.

But the error was that the graph client couldn't be created when using .GetAwaiter()

PR #700, opened 2022-07-25
Add optional parameter to disable welcome message

ohaak2:
Added an optional parmeter to prevent sending of default welcome emails to new members.


ohaak2:
Added filter to GetUnifiedGroup to make sure it's only returning a GroupEntity, if the corresponding group is of type "Unfied". This aligns the method functionality to the GetUnifiedGroups method.

PR #739, opened 2022-09-19
Pnp#727

jekeblad:
Added support for referencing content types in the content type hub


Adam-it:
@jekeblad sorry for the extremely late reply from our side. 
May I kindly ask you to refresh and update a bit the PR description? Could you explain a bit what problem of PnP Provisioning you were trying to solve and how I could test it out?

PR #740, opened 2022-09-19
Added retry when creation status is 3 and the error message is blank

jekeblad:


jansenbe:
@jekeblad : sitestatus 3 means there was an error, 4 means there's a naming conflict...I don't think it's good to keep retrying in those cases. Have you seen provisioning stop early with error 3 but still succeed?

jekeblad:
Hi Bert.
Yes, its strange. Since I never got any error message in the response, I tried again and everything went fine.
I cant remember that it took more than one retry to get it to work I have not seen any side effect of this.

Regards
//Joakim

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows

Issue #331, opened 2021-05-18
List Instance FieldRef for Title Column missing from Site Template


anthonywhite:
### Reporting an Issue or Missing Feature
List Instance FieldRef for Title Column missing from Site Template

### Expected behavior 
ListInstance/FieldRef element should be in template for Title column that is renamed at list level

### Actual behavior 
No FieldRef for Title in ListInstance definition, so column display name is "Title" in a site provisioned from template, rather than display name from source site

### Steps to reproduce behavior

1. Create Site Content Type that includes a Title site column (from Item content type)
2. Add Content Type to a List Instance
3. Rename Title on List Instance Content Type to change Display Name of Title column for that list
4. Run `Get-PnPSiteTemplate` and examine FieldRefs for ListInstance

### What is the version of the Cmdlet module you are running?
1.5.0

### Which operating system/environment are you running PnP PowerShell on?
- [x] Windows
- [x] PowerShell Core 7.1.3

### Workaround
Manually add a FieldRef for the Title column to the ListInstance in the template file with the required DisplayName attribute

anthonywhite:
Keeping alive... anyone disagree?

anthonywhite:
Closing as covered by #776

PR #776, opened 2022-10-25
Add support to exporting renamed Title columns to a PnP template

jackpoz:
Add support to exporting renamed Title columns to a PnP template by not excluding the field when generating a template.
The Title will be included as a FieldRef like below
```xml
<pnp:FieldRefs>
    <pnp:FieldRef ID="fa564e0f-0c70-4ab9-b863-0177e6ddd247" Name="Title" Required="true" DisplayName="Renamed title" />
</pnp:FieldRefs>
```

The field changes are already applied by the provisioning engine so no changes are required there.

I have found some old discussions about this change with different opinions. I hope it will be considered as a good addition.


jackpoz:
Today I found out that this also fixes applying a template with a calculated column that references a renamed Title, where before it was throwing an exception "Your formula is referencing an invalid column" (formulas use the DisplayName)

anthonywhite:
@jackpoz @jansenbe guess this will mean my suggestion #331 can also be closed? 🙂

jackpoz:
I would say so as it implements it

PR #816, opened 2022-12-14
Fix ViewTypeKind property being ignored when provisioning views

jackpoz:
Fix ViewTypeKind property being ignored by SPO when provisioning a new view and updating ListViewXml property during the same ExecuteQueryRetry() call. All views would have ViewType set to HTML before.

The code in this PR first creates the view, calls ExecuteQueryRetry(), then updates the ListViewXml property, then calls ExecuteQueryRetry(). This seems to be needed by serverside behavior of SPO, as ViewCreationInformation is filled with the same information in both cases.

## Tests
The test was done with a view created from SPO List Settings with "Datasheet View" as base view and then exported with Get-PnPSiteTemplate. The xml template has these attributes:
```xml
<View Name="{221F9B40-A0A0-40AE-9AF6-8C109AD3C6B0}" Type="GRID" Personal="TRUE" DisplayName="Grid view" Url="{site}/Lists/TestList/PersonalViews.aspx" Level="1" BaseViewID="1" ContentTypeID="0x" ImageUrl="/_layouts/15/images/generic.png?rev=47">
```
Notice how the view provisioned before the fix has ViewType set to HTML and ListViewXml shows Type="HTML", even if the template has GRID
image

This is how the view looked on SPO, not grid view. A custom sorting by ID was added to test a different case than the standard view one:
image

This is the same view created using the same template after applying the fix from this PR. Notice how ViewType now is set to GRID and ListViewXml has Type="GRID":
image

SPO kept the same sorting by ID and the view is a grid view:
image

In an attempt to test it a bit deeper I created a few views, exported them as pnp template, deleted the list and re-applied the template.
Case with default "All items" plus a new custom field "Category":
image

Case grouping by custom field "Category", sorted by ID descending:
image

Case filtering by custom field "Category" set to "Choice 2":
image

Case with Grid view:
image

Template used for the test cases:
[views.zip](https://github.com/pnp/pnpframework/files/10230443/views.zip)

### DIFF
--- a/src/lib/PnP.Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
+++ b/src/lib/PnP.Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
@@ -764,6 +764,10 @@ namespace PnP.Framework.Provisioning.ObjectHandlers
                 var viewInnerXml = reader.ReadInnerXml();
 
                 var createdView = createdList.Views.Add(viewCI);
+                createdList.Update();
+                web.Context.ExecuteQueryRetry();
+
+                // Edit the view settings after creating it to avoid issues with some creation properties being ignored, for example ViewTypeKind
                 createdView.ListViewXml = viewInnerXml;
                 if (hidden) createdView.Hidden = hidden;
                 createdView.Update();


jackpoz:
@jansenbe just a heads up that I'm getting an error "The target view cannot be configured as a mobile view." in some cases using latest nightly, most likely caused by this PR. I will check and let you know.

jackpoz:
Somehow a pnp template created from an existing site ended up with a Grid view with Mobile view enabled
```xml
<View Name="{08538C78-51BA-4514-9E7D-46F45437F1EC}" MobileView="TRUE" Type="GRID" DisplayName="SomeView" Url="{site}/Lists/SomeList/SomeView.aspx" Level="1" ContentTypeID="0x">
```

No idea how that is even possible to do from the browser (the site was created by users manually). Opening the view from the browser and clicking "Ok" to save it without doing any change disabled the "Mobile view" setting.

Before this PR the view was not created as GRID so the error never showed up. Do you want this to be validated in PnP or should users fix their templates ?

PR #827, opened 2023-01-21
Reduce Memory usage while applying pnp template

stokuri:
Reduce Memory usage while applying pnp template


Adam-it:
duplicate of #825

PR #841, opened 2023-02-17
Returning character ' as invalid

magarma:
Returning back the character ' because since it's valid for URLs, a later call to SharePoint _api/SP.Directory.DirectorySession/Group(alias=....) is failing


magarma:
@jansenbe Thanks for approving yesterday my PR, it seems that despite it was valid, in a later stage this makes failing the provisioning...

PR #850, opened 2023-02-28
Adding code to support multiple default values at folder level.

roberAlb:
Multiple default values for folder were not supported. We have added code to support them for the provisioning.

### DIFF
--- a/src/lib/PnP.Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
+++ b/src/lib/PnP.Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs
@@ -261,7 +261,7 @@ namespace PnP.Framework.Provisioning.ObjectHandlers
                 if (!string.IsNullOrEmpty(fieldValue))
                 {
                     var field = listInfo.SiteList.Fields.GetByInternalNameOrTitle(fieldName);
-                    var defaultValue = field.GetDefaultColumnValueFromField((ClientContext)web.Context, folderName, new[] { fieldValue });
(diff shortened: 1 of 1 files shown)


PedroMordeP:
Hi, I'm getting an error on line `var splitData = terms[q + 1].Split(new char[] { '|' });`

It gives an IndexOutOfRangeException Index was outside the bounds of the array

The scenario is when you have only one value

```
<pnp:Folders>
    <pnp:Folder Name="Contoso Construction">
          <pnp:DefaultColumnValues>
              <pnp:DefaultColumnValue Key="Industry" Value="Construction" />
          </pnp:DefaultColumnValues>
    </pnp:Folder>
<pnp:Folders>
```

What I did at this moment was to check like this

```
        private static List<string> TermIdsToProcess(string value)
        {
            var termDefaultValuesParsed = new List<string>();

            // FIX - Support only one item - to fix the issue on line 'terms[q + 1].Split(new char[] { '|' });' - gives OutOfRange exception
            if (!value.Contains(";#"))
            {
                termDefaultValuesParsed.Add(value);
                return termDefaultValuesParsed;
            }
            // -----

            var terms = value.Split(new[] { ";#" }, StringSplitOptions.None);
            for (int q = 0; q < terms.Length; q += 2)
            {
                var splitData = terms[q + 1].Split(new char[] { '|' });
                var termIdString = splitData[1];
                termDefaultValuesParsed.Add(termIdString);
            }

            return termDefaultValuesParsed;
        }
```

Hope this can help to review/update the code to support this scenario

PR #906, opened 2023-08-09
Added localization for modern audience fields.

nils-a:
Currently, when the "modern audience targeting" fields are added using PnP the display names are not localized. 
When the fields are added using the SharePoint UI, however, they are.
This PR fixed that by adding the correct Resource-Ids as the display names.


jansenbe:
Good improvement @nils-a , thanks for submitting this!

PR #915, opened 2023-09-18
fix(provisioning): non-taxonomy default folder values cause error

JakeStanger:
This fixes a bug in the provisioning engine where default folder values would crash for any field type other than metadata.


jansenbe:
@JakeStanger : nice improvement, thanks for sharing this!

JakeStanger:
I attempted to co-author the commit but I don't think it worked, but credit goes to @fidesimon for finding this and providing an initial fix

PR #951, opened 2023-12-14
Fixing IsAvailableForTagging not being set correctly on a reused term

KoenZomers:
Fixing IsAvailableForTagging not being set correctly on a reused term. It would always use the default of being true, even though the XML import would define it needing to be false. This fixes that it will apply whatever is configured in the XML.

image


gautamdsheth:
Thanks @KoenZomers !

Issue #963, opened 2024-01-15
Question regarding "TokenParser" because of pending optimizations


fzbm:
Hello PnP team,

I have some question regarding some parts of the `TokenParser`.

Those question arose when I analyzed and optimized the performance of the `TokenParser`, because PnP behaved really poorly in the performance department when I tried to deploy a fairly large template. It took unusually long to deploy a field, a content type or a list (up to minutes). Also the initialization of the `TokenParser` alone took over **10 minutes** on an **S2** App Service Plan and **1.5 minutes** on an **i7-12800HX**. Needless to say, that the CPU usage was really high.

With the usage of a profiler I was able to pinpoint the initialization issue to how the localization resource tokens were added to the parser. The template in questions defined a larger amount of languages and therefore also a larger amount of resource tokens. At the end the parser had about **48,000** definitions and above **300,000** tokens.

A second issue was the (re)building of the cache, which happened every time a string should be parsed. I know that a token gets skipped if he is already a part of the cache, but iterating through all definitions and their tokens alone in combination with using `Regex.Unescape` every time to then just skip it, is really taxing the CPU. This was also something the profile has shown.

I already made changes to the `TokenParser` and was able to reduce the initialization time on my machine from about 1.5 minutes to **35 seconds**. The processing of the resource tokens alone required 1 minute and has been reduced to **about a second**. By the way the culprit regarding the resource tokens was the LINQ operation looking for matching entries at the end.

I will create a pull request and link this issue to it. I might also create additional commits based on the outcome of this discussion.

<ins>**Now my questions.**</ins>

[`Rebase(Web)`][1] only calls `ClearCache` on every definition, but does not clear the actual cache of the `TokenParser`. Therefore clearing the cache of a `TokenDefinition` would only affect non-cacheable definitions. Is this an intended behavior?

---

[`AddToken`][2] and the [constructor accepting a web, a template and apply information][3] always sort the token definitions based their longest token in a descending order. Is this still required? If I haven't overlooked something, the order is not important for any of the operations. Is my assumption correct and the sorting could be removed?

This is a source for a high CPU utilization. Especially when performed every time a token gets added by using `AddToken`.

If the order is still required, adding a method like `AddTokens` would be a solution. It would allow adding multiple tokens at once. The code of the `ObjectHandler` adding tokens must be changed to add all at once (if possible).

---

[`RemoveToken`][4] does not update the `ListTokenDictionary` dictionary. Is this intended or an oversight?

---

[...]

PR #964, opened 2024-01-16
Optimizations to "TokenParser"

fzbm:
This Pull Requests makes changes to the `TokenParser` to optimize the performance when working with larger templates. There is the companion ticket #963 which also contains some questions, which might result in additional commits once clarified. The `Clone` topic is one of those.

As described in the ticket the performance of PnP was really bad when deploying a larger template which also contained a larger amount of localization resources. It took several minutes to "initialize the engine" (aka the `TokenParser`) and also up to minutes when deploying a field, content type or list, but the problem has been already described in greater detail in the ticket #963.

With the help of a profiler I was able to pinpoint the source of the problem to the `TokenParser`, which really taxed the CPU when processing the localization resources and also when parsing a string (`ParseString`) due to repeatedly iterating through all `TokenDefinition`, their tokens and processing them with `Regex.Unescape`. I made some changes to the initialization of the localization resource tokens and the parsing of strings, which greatly reduced the initialization time of the `TokenParser` and also the overall CPU utilization when working with `ParseString`.

The following main changes have been made.

* The token cache now only gets created when using `Rebase`, the private constructor (for cloning) and creating the instance.
* Added a cache for non-cacheable `TokenDefinition` containing the actual `TokenDefinition` instead of the value. This has been added to eliminate the need of creating a dictionary with all non-cacheable definitions most of the times when `ParseString` gets called.
* The normal and non-cacheable defintions cache gets updated by just processing the affected `TokenDefinition` when calling `AddToken` or `RebuildListTokens`. In most cases the cache dictionaries are initialized with a calculated capacity to avoid resizing operations.
* Changed `AddResourceTokens` to use a dictionary instead of a list to avoid costly lookup operations when adding `LocalizationToken`. Additionally the capacity of the `_tokens` list gets extended to a calculated amount beforehand to avoid resizing operations.
* `GetListTitleForMainLanguage` and `_listsTitles` are now instance members. This fixes issues with environments where multiple threads might deploy a template at the same time.
  * `GetListTitleForMainLanguage` now also uses a batch approach for retrieving the titles of the lists in order to reduce the time and amount of server calls required to retrieve them.
* Removed the sorting of `_tokens` when adding a token or creating a `TokenParser` instance. It looks like sorting the list is not/no longer required, but might be added back later based on the discussion in ticket #963.

The following additional changes have been made. The first two are not part of any of the commit messages due to an oversight on my side.

* **ResourceEntry**
  * Changed the type of the `LCID` property from `uint` to `int`.
* **LocalizationToken**
  * Changed `LocalizationToken` to use a `Dictionary` instead of a `List` to speed up the lookup performance when trying to retrieve a `ResourceEntry` for a specific language.
    * The `ResourceEntries` property is now also an `IReadOnlyList` instead of a `List`. A caller should not be able to at least add or remove a `ResourceEntry` on an `LocalizationToken` instance. He would still be able to manipulate the `ResourceEntry` itself though. The change should not have any impact on existing applications, because the whole class is `internal` and not available outside of the library.
* **TokenParser**
  * Formatted some code to enhance the readability.
  * Replaced some LINQ methods with native methods provided by the specific collection type.
  * Changed the code of some methods to exit them early, which reduces the nesting.
  * Removed unused parameters from private methods.
  * Removed some commented code.
  * Moved all variable declarations to the top of the class.

### DIFF
--- a/src/lib/PnP.Framework/Extensions/ClientContextExtensions.cs
+++ b/src/lib/PnP.Framework/Extensions/ClientContextExtensions.cs
@@ -1,18 +1,9 @@
-using PnP.Framework;
-using PnP.Framework.Diagnostics;
-using PnP.Framework.Http;
-using PnP.Framework.Provisioning.ObjectHandlers;
-using PnP.Framework.Sites;
-using PnP.Framework.Utilities;
-using PnP.Framework.Utilities.Async;
-using PnP.Framework.Utilities.Context;
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Configuration;
 using System.Diagnostics;
 using System.IdentityModel.Tokens.Jwt;
-using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
@@ -21,6 +12,14 @@ using System.Text.Json;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using System.Xml;
+using PnP.Framework;
+using PnP.Framework.Diagnostics;
+using PnP.Framework.Http;
+using PnP.Framework.Provisioning.ObjectHandlers;
+using PnP.Framework.Sites;
+using PnP.Framework.Utilities;
+using PnP.Framework.Utilities.Async;
+using PnP.Framework.Utilities.Context;
 
 namespace Microsoft.SharePoint.Client
 {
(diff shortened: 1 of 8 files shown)


fzbm:
Are there any updates about this? This is somewhat a production impacting issue for an already active application and heavily impacts a planned feature.

jansenbe:
@fzbm : this is pretty big change requiring a lot of test effort as token parser is fundamental to the provisioning engine. Given I can only spare a few cycles a month on this project I currently don't have the bandwidth to extensively test this. Can you explain more about this: are you using this in your application today, do you feel your provisioning templates are representative of what folks typically would do and do they still work fine using your change?

fzbm:
A small update from my side.

The modified version is in production for about two days and deployed since then 24 different templates to 97 sites. We have observed no issues so far, but what we definitely noticed is a huge improvement regarding the performance.

There are times were it happens that 10 concurrent jobs would deploy templates to 10 different sites. Before the changes situations like this would lead to a 100% CPU load and a really long runtime time of each job. With those changes the average load in the same situation is about 40 to 45% percent and the time is about the same as if only one job was executed. In both cases the code was running on a P1v2 App Service.

Logically we also noticed a runtime reduced by half in some cases when deploying a template, although this again depends on the template in question.

NicolajHedeager:
This is definitely a lot more efficient than the existing TokenParser.
Have been trying out this implementation for a few weeks, no issues so far.

Issue #966, opened 2024-01-18
c# ApplyProvisioningTemplate: Unable to load one or more of the requested types


AlyaKoni:
Azure Function in Visual Studio 2022
```
Azure Functions Core Tools
Core Tools Version: 4.0.5455 Commit hash: N/A (64-bit)
Function Runtime Version: 4.27.5.21554
```

Framework and Libraries:
```
<TargetFramework>net6.0</TargetFramework>
<AzureFunctionsVersion>v4</AzureFunctionsVersion>

<PackageReference Include="Azure.ResourceManager" Version="1.10.0" />
<PackageReference Include="Azure.ResourceManager.Network" Version="1.7.0" />
<PackageReference Include="HtmlAgilityPack" Version="1.11.57" />
<PackageReference Include="Azure.Storage.Files.Shares" Version="12.17.1" />
<PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Http" Version="3.2.0" />
<PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="5.2.2" />
<PackageReference Include="Microsoft.Graph.Beta" Version="5.62.0-preview" />
<PackageReference Include="Microsoft.IdentityModel.JsonWebTokens" Version="7.2.0" />
<PackageReference Include="Microsoft.NET.Sdk.Functions" Version="4.2.0" />
<PackageReference Include="Azure.Data.Tables" Version="12.8.2" />
<PackageReference Include="Azure.Security.KeyVault.Certificates" Version="4.5.1" />
<PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.5.0" />
<PackageReference Include="Azure.Storage.Blobs" Version="12.19.1" />
<PackageReference Include="Azure.Storage.Queues" Version="12.17.1" />
<PackageReference Include="Microsoft.SharePointOnline.CSOM" Version="16.1.24322.12000" />
<PackageReference Include="PnP.Framework" Version="1.14.0" />
<PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="7.2.0" />
```

Trying to copy a template from one site to other:
```
template = srcContext.Web.GetProvisioningTemplate(ptci);
dstContext.Web.ApplyProvisioningTemplate(template, ptai);
```

Getting an exception:
```
Unable to load one or more of the requested types.
[...]

PR #967, opened 2024-01-18
Fixed GUID strings

AlyaKoni:
Guid strings are wrong. Dozens have a missing '}' at the end.


AlyaKoni:
Added more robust type loading to fix issue #966

src/lib/PnP.Framework/Provisioning/Providers/Xml/XmlPnPSchemaBaseSerializer.cs
```diff
@@ -683,8 +683,19 @@ public ProvisioningHierarchy ToProvisioningHierarchy(Stream hierarchy)
...
+            {
+                serializerTypes = currentAssembly.GetTypes();
+            }
+            catch (ReflectionTypeLoadException e)
+            {
+                serializerTypes = e.Types.Where(t => t != null).ToArray();
+            }
+
+            // Get all the serializers
+            var serializers = serializerTypes
```
nils-a, line 698: While I'm personally not affiliated with maintaining this repository in any way, I feel this change is a bad one.

The `BuildInFieldId` in the current state is fundamentally broken, and this change only obscures that fact by hiding the error.

AlyaKoni: My pull request fixed two issues. Firstly, the fundamentally wrong strings that break the BuildInFieldId. Secondly, a type loading problem in azure functions. Please keep these two problems apart.

nils-a: I see. I thought those problems were connected.

jansenbe:
@AlyaKoni : thanks for submitting this fix. We've decided to go for the alternative fix proposed by @nils-a which should also get your use case unblocked.

Issue #968, opened 2024-01-19
Error when calling Get-PNPSiteTemplate: The type initializer for 'PnP.Framework.Enums.BuiltInFieldId' threw an exception.


nils-a:
On the _nightly_ version 2.3.29 and 2.3.30, when calling `Get-PNPSiteTemplate` - on everything that includes `Fields`, the operation fails with a message:

> Get-PnPSiteTemplate: The type initializer for 'PnP.Framework.Enums.BuiltInFieldId' threw an exception.

When the TraceLog is set to `Debug`, there are actually none (in numbers: 0) lines logged.

This error **does not** appear on the current 2.3.0 version.

nils-a:
The error seems to originate in the `PnP.Framework.Enums.BuiltInFieldId.Contains` method. (Which is called for every field found in `PnP.Framework.Provisioning.ObjectHandlers.ObjectField.ExtractObjects`.)

My guess is, that https://github.com/pnp/pnpframework/pull/967 will fix this issue, too. 😀

emdeka:
The following GUIDs in https://github.com/pnp/pnpframework/blob/dev/src/lib/PnP.Framework/Enums/BuiltInFieldId.cs contain  invalid GUIDs, note the missing bracket closing the ID. Starting in:

```
public static readonly Guid SkipFeatureDeployment = new Guid("{eda244c9-5a10-4057-8b96-a118a3b31454");
public static readonly Guid AppPackageErrorMessage = new Guid("{80ac5bc8-73dc-4967-950f-3f6ec32c19f4");
public static readonly Guid AppPermissionXML = new Guid("{bad900c5-34c6-4d4b-b84a-698b5da331fe");
public static readonly Guid AppSubtype = new Guid("{de81f63e-c919-4f47-80fb-95ef0ef506f7");
```

And stopping in:

```
public static readonly Guid WebApiPermissionScopesNote = new Guid("{5d88909c-873a-4262-bc81-9b7c2ad04105");
public static readonly Guid TenantWideExtensionWebTemplate = new Guid("{8b97c8ba-e0c8-4c35-9b65-efee64f1b610");
public static readonly Guid WebApiPermissionRequests = new Guid("{f8a04db8-4b6e-4613-9232-02811628a7f6");
```

Sorry, I don't know much about git to help here, but maybe someone else can fix this.

PR #972, opened 2024-01-24
fix missspelled Guids because of Error

czullu:
Guid should contain 32 digits with 4 dashes (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx)


czullu:
#968 is the better fix as it contains the tests as well..

czullu:
#968 is the better fix as it contains the tests as well..

PR #1016, opened 2024-04-25
fixing issue #1015

### LINKED ISSUE #1015: Web extension DisableRequestAccess not working
I want to disable "__Allow Access requests__" in Site sharing settings. I mean this in the UI:
image

There is a Web extension that should do the job (and worked in the past) but is not working now. This is the current code:

```csharp
        public static void DisableRequestAccess(this Web web)
        {
            web.RequestAccessEmail = string.Empty;
            web.Update();
            web.Context.ExecuteQueryRetry();
        }
```

As per my tests now is not enough with setting an Empty email and an extra step is required. 

I'm about to do a PR to fix it, but openning an issue first for traceability.

luismanez:
Should fix #1015 

In the past, the method was working fine with just setting an Empty value in the ```RequestAccessEmail```. However, this is not working anymore (I tested it in 2 different tenants).
Now it needs an extra step first calling ```SetUseAccessRequestDefaultAndUpdate(false)```


gautamdsheth:
Thanks @luismanez , have merged it !!

PR #1030, opened 2024-06-14
fix(auth): fix an issue where an asp.net application could potentially freeze when targetting a SharePoint On-premise environment

nathan-swannet:
In the ConfigureOnPremisesContext an event handler is registered containing an async function which is invoked with .GetAwaiter().GetResult();
This can and will freeze your whole asp.net application due to the nature of how the continuation of that task is done in this context.

Wrapping it in a Task.Run fixes this issue and ensures any exception are still thrown in the current synchronisation context.


gautamdsheth:
Thanks @nathan-swannet , sorry for the delay in this !

PR #1070, opened 2024-10-09
Giving support for Everyone except external users during the ListItem data row extraction

magarma:
Currently when extracting list instance data rows for user fields, the user "Everyone except external users" is ignored.
This commit tries to give support for it.
After this, it will look like this:

<pnp:DataValue FieldName="CustomField">alaynawhite@whatever.onmicrosoft.com,Knowledgegroup@whatever.onmicrosoft.com,AnotherGroup@whatever.onmicrosoft.com,{everyonebutexternalusers}</pnp:DataValue>


gautamdsheth:
Thanks @magarma  !!

PR #1079, opened 2024-10-28
Fix #1078 by improving error handling

### LINKED ISSUE #1078: [BUG] ClientContext.IsAppOnly() might fail when authenticated with ClientCredentials
### Reporting an Issue or Missing Feature

There's an issue when you add ClientContext.IsAppOnly() and are using ClientCredentials. It seems to randomly throw an uncaught 403 and crash.

### Expected behavior 
When calling .IsAppOnly() it should return true or false, not throw (at least not if it is able to fetch the Access Token!)

### Actual behavior 
When debugging it, I noticed it doesn't happen on first execution of the handler for extracting the AccessToken, but rather on one of the subsequent ones. It's not even actually executing a real query (it's a dummy) but that still fails somehow.

1. Connect-PnPOnline with ClientCredentials
2. Call ClientContext.IsAppOnly() in the commandlet's code
3. Observe

### What is the version of the Cmdlet module you are running?
Latest dev (f3b8aeffdccc168355d8335ef17d884fd20235b1)

### Which operating system/environment are you running PnP PowerShell on?
- [X] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

### Next steps

I'll have a fix proposed in a minute.

koskila:
Don't throw in ClientContext.IsAppOnly() if AccessToken extraction is successful.

### DIFF
--- a/src/lib/PnP.Framework/Extensions/ClientContextExtensions.cs
+++ b/src/lib/PnP.Framework/Extensions/ClientContextExtensions.cs
@@ -576,7 +576,6 @@ namespace Microsoft.SharePoint.Client
         /// <returns>True if app-only, false otherwise</returns>
         public static bool IsAppOnly(this ClientRuntimeContext clientContext)
         {
-
             // Set initial result to false
             var result = false;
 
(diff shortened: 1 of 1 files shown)


gautamdsheth:
Thanks @koskila , merged it !!

PR #1081, opened 2024-10-30
Handle urls ending with / when creating new sites

jackpoz:
Handle urls ending with / when creating new sites by removing any empty entry after splitting the url by /.
This is needed when the url ends with / as the last part after `Split('/')` will just be an empty string, causing `Replace("", "")` to be called, throwing an exception.

This has been reported at https://github.com/pnp/powershell/issues/4491 and can be reproduced calling New-PnPSite:
```powershell
 New-PnPSite -Type TeamSiteWithoutMicrosoft365Group -Owner <owner upn> -Url "https://<tenant>.sharepoint.com/sites/TestWithSlash/" -Title "Test With Slash"
```
image

The PR fixes the issue and the site is created:
image


gautamdsheth:
Thanks @jackpoz , merged it !

PR #1084, opened 2024-11-08
General channel is not created by default anymore, check first channel when creating and initialize Teams

magarma:
General channel is not created by default anymore.
Now Teams allows to define the first Channel name when creating the Teams.
Initialize the first channel found.


gautamdsheth:
Thanks @magarma , merged it !

PR #1124, opened 2025-04-03
Support for Federated identity credential authentication

JGemkow:
Adding support for authentication via a combination of user-assigned managed identities and [federated identity credentials](https://learn.microsoft.com/en-us/graph/api/resources/federatedidentitycredentials-overview?view=graph-rest-1.0). With this approach, you can configure a user-assigned managed identity to be a federated identity credential on an Entra ID application registration. An application running in Azure can obtain a token using its user-assigned MSI and then exchange that for a token for the application registration.

This has been tested for certificateless/secretless cross-tenant access to SharePoint. For example, you can create a multi-tenant application registration in Tenant A and add the user-assigned managed identity of an Azure Function in that same tenant to be a Federated Identity Credential on that app registration. On that application registration, you can add application permissions for Microsoft Graph and SharePoint's _Sites.Read.All_ role. Should you need to access data in another target tenant, you can instantiate a service principal in that target tenant (Tenant B) and have an application/global admin consent. Once complete, your code running in the Azure Function in Tenant A can leverage the new AuthenticationManager ` CreateWithManagedIdentityFederatedIdentityCredential` method for setting up authentication and then getting authenticated client contexts for use with CSOM all without having to rotate/manage certificates and secrets.

This uses a pattern described here: https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation-config-app-trust-managed-identity?tabs=microsoft-entra-admin-center#msal-net


gautamdsheth:
Thanks @JGemkow , have merged this !!

Issue #1156, opened 2025-07-28
PnP.Framework.Graph.UsersUtility.ListUserDelta() issue with null deltaToken - Value cannot be null. (Parameter 'input')


svermaak:
@gautamdsheth, I believe https://github.com/pnp/pnpframework/commit/676399b44ed319f29f3e623ffd0e2b47571c79a4 or prior broke PnP PowerShell's ability to run this

`$newAllUsers = Get-PnPAzureADUser -Connection $tenantConnection -Delta`

All of a sudden this does not work if delta token is null (which is what is expected from the above command. It is how the first run is done to get skip token for future runs)
```
public static UserDelta ListUserDelta(string accessToken, string deltaToken, string filter, string orderby, string[] selectProperties = null, int startIndex = 0, int? endIndex = null, bool useBetaEndPoint = false, AzureEnvironment azureEnvironment = AzureEnvironment.Production)
{
    var userDelta = PnP.Framework.Graph.UsersUtility.ListUserDelta(accessToken, deltaToken, filter, orderby, selectProperties, startIndex, endIndex, useBetaEndPoint: useBetaEndPoint, azureEnvironment: azureEnvironment);

    var result = new UserDelta
    {
        DeltaToken = userDelta.DeltaToken,
        Users = userDelta.Users.Select(User.CreateFrom).ToList()
    };
    return result;
} 
```

PR #1157, opened 2025-07-28
Update UsersUtility.cs

svermaak:
This PR introduces a change to the ListUserDelta method to ensure that the $skiptoken query parameter is only included in the request if the deltaToken parameter is not null or empty. This change addresses scenarios where the deltaToken is not provided, ensuring the query string remains valid and avoids unnecessary parameters.

#1156


gautamdsheth (approved):

gautamdsheth:
Thanks @svermaak , merged it !

PR #1168, opened 2025-08-29
RequestStream creation before refering to it

ironNPNcake:
After a change in AuthenticationManager, which is using ClientContext extension to add WebRequestExecutorFactory, HTTP requests are failing with null reference exception, as _requestStream was never initialized and .Seek method is failing to execute. Added instance creation before executing seek method


gautamdsheth:
Thanks @ironNPNcake , merged it !

Issue #1175, opened 2025-09-22
Provisioning engine does not export the Collapsible or DisplayName properties of page sections


johnnycardy:
Extract a page with a collapsible section - the Collapsible property in the XML template is not set.

The provisioning is already implemented, so setting it to true manually will correctly result in a collapsible section.

johnnycardy:
Ah, already fixed.

PR #1176, opened 2025-09-22
Extract Collapsible and DisplayName properties of page sections

johnnycardy:
Currently, extracting pages does not extract the "DisplayName" or "Collapsible" property of sections.

See issue #1175


johnnycardy:
Already fixed in dev branch - oops.

PR #1177, opened 2025-09-24
Fix comparison logic for list field settings supporting export/import Autofill Info

magarma:
Support exporting fields with autofill info in list.
image


gautamdsheth:
Thanks @magarma , merged this !

PR #1182, opened 2025-10-06
Fix for IsCurrentUserTenantAdmin if the user is not a tenant admin.

patrikhellgren:
Before this PR if the user was not a tenant admin an exception would be thrown. This changes the behaviour so that it returns false like it should.


gautamdsheth:
Awesome stuff @patrikhellgren !

PR #1003, opened 2024-04-01
Remove graph api sdk dependency

quails4Eva:
Reduce dependency on Microsoft.Graph NuGet package
Based on discussion in  #660 
https://github.com/pnp/pnpframework/issues/660#issuecomment-1135454556

PR #1183, opened 2025-10-09
Fix Graph group serialization.

patrikhellgren:
Fixes a Graph Group serialization issue introduced in #1003.
Without this a 400 Bad Request was returned from Graph when creating a new Group, e.g. during provisioning with app credentials.


gautamdsheth:
Thanks @patrikhellgren  !!

PR #1184, opened 2025-10-09
Catch unauthorized exception when getting tenant themes.

patrikhellgren:
In some cases the app credentials would pass the tenant admin check but still not have access to the tenant themes and an unauthorized exception would be thrown, e.g. with Group.ReadWrite.All and Group.Create permissions.


gautamdsheth:
Thanks @patrikhellgren !!

PR #1185, opened 2025-10-11
Fix for getting the unified group id efter creating a new group

patrikhellgren:
Fixes an issue introduced in #1003.
The response from creating a new unified group has the group id in a different property.


gautamdsheth:
Thanks @patrikhellgren

PR #1209, opened 2025-12-30
Update path handling for cross-platform compatibility

patrikhellgren:
Replaced hardcoded path separators with Path.DirectorySeparatorChar and Path.Combine in provisioning file system connector and file utilities.
This fixes an issue where the provisioning file handler would fail when recursively copying files on Linux.


gautamdsheth:
Thanks @patrikhellgren , merged it !

PR #1210, opened 2025-12-30
Increase retry delay and improve error handling in GetTabs

patrikhellgren:
Fixes an issue with provisioning tabs to a newly created channel which sometimes would not be found until retrying a few times, especially for private channels.


gautamdsheth:
Thanks @patrikhellgren , merged it !

Issue #1212, opened 2026-01-07
Collapsible headingLevel not set


czullu:
While the headingLevel is correctly exported to the Template-XML, it is not set during provisioning 

<img width="1662" height="632" alt="Image" src="https://github.com/user-attachments/assets/57c4b49b-ea7f-415d-b395-eae57daacc57" />

Working on Fix

PR #1213, opened 2026-01-07
extend CanvasSection Model with HeadingLevel for collapsible-section-tilte

czullu:
as XML-Template does not support this property, we get it from Control.JsonControlData -> zoneGroupMetadata.headingLevel in case it's not specified in a other way through code


czullu:
Fix for #1212

gautamdsheth:
@czullu - this requires schema changes, not sure we can merge it as it is in this state.

PR #1216, opened 2026-01-15
Ensure List Fields Are Included When Site Column Titles Differ

roberAlb:
-Load and compare the Title property for site columns in ObjectListInstance.cs to ensure list fields are included when titles differ from what is coming at site column.


gautamdsheth:
Thanks @roberAlb , merged it !!

PR #1247, opened 2026-07-03
Fix broken add-owner/add-member Graph calls in Groups utilities

singingknight:
UpdateOwners/UpdateMembers in both GroupsUtility and UnifiedGroupsUtility passed accessToken positionally into MakePostRequest parameter 2, which is the request body (object content) - not the access token. This sent the token as the POST body with no auth header and no @odata.id reference payload, so adding owners/members was non-functional. The bug came from the Graph SDK -> HttpHelper migration (2435d060, cf619edd).

Align all four add/remove paths with the working ObjectTeams pattern:
- POST .../{owners|members}/$ref with an {"@odata.id": ".../users/{id}"} body, application/json content type, and accessToken in the correct slot
- Correct the DELETE reference URLs from /ref to /$ref

### DIFF
--- a/src/lib/PnP.Framework/Graph/GroupsUtility.cs
+++ b/src/lib/PnP.Framework/Graph/GroupsUtility.cs
@@ -121,8 +121,9 @@ namespace PnP.Framework.Graph
                     try
                     {
                         // And if any, add it to the collection of group's members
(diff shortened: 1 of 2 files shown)


gautamdsheth:
Thanks a lot @singingknight for fixing this, much appreciated !!

Issue #1230, opened 2026-04-15
List FieldRef (Ref to Site Column)  won't be exported if this list enabled content type management and this field is not visible on list content types


wuxiaojun514:
I found an interesting problem when run `Export-PnPListToSiteTemplate` command.

# My Scenario
Let me describe my scenario first.

I have one list which enabled content type management, because I have two content types under this list 

<img width="604" height="143" alt="Image" src="https://github.com/user-attachments/assets/a258b072-283c-469d-a2a8-7069322c16af" />

Some of the list fields are site column

<img width="1338" height="751" alt="Image" src="https://github.com/user-attachments/assets/81cd0a07-b9bc-4538-b499-61e6719fdba0" />

And these two site columns are set as hidden in my list content type

<img width="1416" height="726" alt="Image" src="https://github.com/user-attachments/assets/0e6e4aa0-66d3-4872-91fe-8da64740148e" />

# Missing site column field ref in the exported template
Then it (those two site column field ref) won't get exported when I run  `Export-PnPListToSiteTemplate` or other export template command.
In this case it only exported "Title" since "Tifle" field is visible in list content type
<img width="1133" height="448" alt="Image" src="https://github.com/user-attachments/assets/0f250ecc-5b86-4b27-ba21-69f44fdd0664" />

However if I disable the content type management on the list, then I can get all site columns used in this list.

<img width="1347" height="596" alt="Image" src="https://github.com/user-attachments/assets/c9f8da57-34b5-46ab-bd93-1c7992d505a2" />

In the source code https://github.com/pnp/pnpframework/blob/dev/src/lib/PnP.Framework/Provisioning/ObjectHandlers/ObjectListInstance.cs#L2776, it will not add hidden site column as field ref if the list enabled content type 
``` JavaScript 
  if (siteList.ContentTypesEnabled && contentTypeFields.FirstOrDefault(c => c.Id == field.Id) == null)
                    {
                        if (contentTypeFields.FirstOrDefault(c => c.Id == field.Id) == null)
                        {
                            addField = false;
                        }
                    }
```

wuxiaojun514:
I am not very clear why we put this logic to stop export site column field ref when enabled content type, but obviously the `Invoke-PnPSiteTemplate`  will fail if **pnp:FieldRef** missing.  

I can try to raise a pr for it.

Adam-it:
@wuxiaojun514 interesting find. 
Please do rise a PR for this if possible 👍 
[...]

PR #1249, opened 2026-07-10
fix for issue 1230.

wuxiaojun514:
# Hotfix for issue #1230 

## Problem

Currently I find an issue in list provisioning:

If the list has custom content type which contains hidden site column , the exported template won't have this field ref . See detail in #1230 

It will become a problem if its site level content type doesn't have this field (only list level content type has this field ), then this field won't be provisioned.

## My hotfix

I checked the related code and found that this module is quite old and stable (It doesn't have too much change compared with its initial version)

It seems that the first author want to exclude the system fields (system fields won't be added into content type)  in the list provisioning.

Now `contentTypeFields` will add all the field link in the list content types no matter whether it is hidden or not, to avoid this issue.


Adam-it (approved):
LGTM 🔥

PR #1250, opened 2026-07-11
Updates provisioning templates. Closes #1242

### LINKED ISSUE #1242: Base Templates of Provisioning Engine require an update
The base templates of the PnP Provisioning Engine require an update, as they are heavily outdated and refer for instance to deprecated features.

Adam-it:
## 🎯 Aim

The aim is to refresh base provisioning templates to refresh with latest updates done in SPO.

## Problems

When generating the templates, I had to perform minor code tweaks which is rather not problematic but the bigger issue is that I was unable to create all sites base on the templates we support. In some cases I got an error that this kind of template is not available in my tenant, like the classic site blog.

This is the list of updated templates 

image


Adam-it:
@pkbullock since we are under migration of PnP Provisioning to PnP.Core.Provisioning I would not focus on this PR for now

Issue #1252, opened 2026-07-30
Missing ContentTypeId when extracting list and library data from a sub site


gszdev:
While extracting a site template from a SharePoint sub site content types of the root web are not present in a sub sites content types collection.
So extracting list items / library content (folders, files) cannot be associated with the right ContentTypeId because the content type comes from the root web. 

PnP.Framework 1.20.x

PR #1253, opened 2026-07-30
Fix for #1252 - Missing ContentTypeId when extracting list and library data from a sub site

gszdev:
Fix for #1252


Adam-it (approved):
LGTM 🔥

PR #1257, opened 2026-07-31
Make the provisioning field XML cache thread safe

### LINKED ISSUE #1243: [BUG] Concurrent provisioning can corrupt field XML cache
When multiple provisioning operations run concurrently in the same .NET process, `FieldAndListProvisioningStepHelper` can throw:

```text
System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access.
A concurrent update was performed on this collection and corrupted its state.
   at System.Collections.Generic.Dictionary`2.set_Item(TKey key, TValue value)
   at PnP.Framework.Provisioning.ObjectHandlers.FieldAndListProvisioningStepHelper.GetFieldProvisioningStep(...)
   at PnP.Framework.Provisioning.ObjectHandlers.ObjectField.ProvisionObjects(...)
   at Microsoft.SharePoint.Client.WebExtensions.ApplyProvisioningTemplate(...)
```

The cause appears to be this process-wide static cache:

```csharp
static readonly Dictionary<Field, XElement> _fieldXmlDictionary = new Dictionary<Field, XElement>();
```

Because the dictionary is `static`, independent provisioning runs share it. They do not need to share the same `Field` instance; concurrent writes with different keys are enough to corrupt a normal `Dictionary`.

### Expected behavior

Independent provisioning operations should be able to run concurrently in the same process without corrupting shared PnP Framework state.

### Actual behavior

Concurrent `ApplyProvisioningTemplate(...)` executions can write to `_fieldXmlDictionary` at the same time and throw `InvalidOperationException`. In long-running hosts, such as Azure Functions, the corrupted static dictionary may affect later provisioning runs until the process restarts.

Run multiple `ApplyProvisioningTemplate(...)` calls concurrently in the same process with templates containing fields or list fields. This can happen in worker services, web jobs, Azure Functions, durable activities, or other hosts that process multiple provisioning jobs in parallel.

### What is the version of the library you are running?

Observed with a fork based on PnP Framework `1.19.x`. The same static dictionary exists in the current `dev` branch:

```text
src/lib/PnP.Framework/Provisioning/ObjectHandlers/FieldAndListProvisioningStepHelper.cs
```

### Which operating system/environment are you running on?

.NET isolated Azure Functions app on .NET 10, running multiple durable activity functions concurrently in the same worker process.

This is not Azure Functions specific; any multi-threaded host can hit the same shared static dictionary.

### Proposed solution

Replace the static `Dictionary<Field, XElement>` with `ConcurrentDictionary<Field, XElement>` and use `GetOrAdd` when parsing/caching field XML.

This keeps the current cache behavior and API surface unchanged while preventing concurrent writes from corrupting the shared cache.

```csharp
static readonly ConcurrentDictionary<Field, XElement> _fieldXmlDictionary = new ConcurrentDictionary<Field, XElement>();

private static XElement GetSchemaElement(Field templateField, TokenParser parser, params string[] tokensToSkip)
{
    return _fieldXmlDictionary.GetOrAdd(
        templateField,
        field => XElement.Parse(parser.ParseXmlString(field.SchemaXml, tokensToSkip)));
}
```

svermaak:
## Type of change

- [x] Bug fix

## Related issue

Fixes #1243

`FieldAndListProvisioningStepHelper` cached parsed field schema XML in a process wide static `Dictionary<Field, XElement>`:

```csharp
static readonly Dictionary<Field, XElement> _fieldXmlDictionary = new Dictionary<Field, XElement>();
```

The three helpers that read it, `GetFieldProvisioningStep`, `GetFieldId` and `GetSchemaXml`, each did an unsynchronised check-then-insert on that dictionary. Because the cache is static, provisioning runs that are independent from the caller's point of view still share it, so two `ApplyProvisioningTemplate` calls running at the same time in one process write to the same `Dictionary` concurrently and corrupt it. That surfaces as

```text
System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access.
```

thrown out of `ObjectField.ProvisionObjects`. The keys do not need to collide for this to happen, and a corrupted `Dictionary` can also spin inside a lookup, so a long running host can keep failing until the process is recycled.

This replaces the cache with a `ConcurrentDictionary` and routes all three helpers through one `GetOrAdd` based `GetCachedSchemaXml`, which also removes the three copies of the same check-then-insert block. Caching behaviour, the keys used and every method signature stay the same.

### Verification

A small console harness drives the real code path: 8 concurrent runs, each with its own template fields, 400 fields per run, classified through `GetFieldProvisioningStep` exactly as `ObjectField` does. Nothing is shared by the caller, so the only shared state is the framework's static cache.

Before, every run fails with the exception from the issue:

Before the fix

After, all 3200 fields are classified without error, and a follow up check confirms the cache still serves repeated fields and still routes `Lookup`/`Calculated` fields to `Step.LookupFields`:

After the fix

Builds clean for `netstandard2.0`, `net8.0` and `net9.0`.

### Note

I could not add a regression test to `PnP.Framework.Test`: it targets `net10.0` only and I have no .NET 10 SDK on this machine, so anything I wrote there would go in unverified. The harness above needs reflection anyway, since these helpers are `internal` and `TokenParser`'s token list constructor is private. Happy to add a test in the shape you prefer if you would like one.

Two related things I deliberately left alone, since they are separate from the crash in the issue:

- The cache is never trimmed, so it grows for the lifetime of the process.
- `Field.GetHashCode()` parses the schema XML on every call, so each cache lookup already costs one `XElement.Parse`, which is most of what the cache is meant to avoid.

No CHANGELOG entry included.

### DIFF
--- a/src/lib/PnP.Framework/Provisioning/ObjectHandlers/FieldAndListProvisioningStepHelper.cs
+++ b/src/lib/PnP.Framework/Provisioning/ObjectHandlers/FieldAndListProvisioningStepHelper.cs
@@ -1,5 +1,5 @@
 ﻿using System;
-using System.Collections.Generic;
+using System.Collections.Concurrent;
 using System.Xml.Linq;
 using Field = PnP.Framework.Provisioning.Model.Field;
 
@@ -7,15 +7,13 @@ namespace PnP.Framework.Provisioning.ObjectHandlers
 {
     public static class FieldAndListProvisioningStepHelper
     {
-        static readonly Dictionary<Field, XElement> _fieldXmlDictionary = new Dictionary<Field, XElement>();
+        // Process wide cache, so it is shared by every provisioning run in the process.
+        // A plain Dictionary corrupts itself when two runs provision fields at the same time.
+        static readonly ConcurrentDictionary<Field, XElement> _fieldXmlDictionary = new ConcurrentDictionary<Field, XElement>();
+
         internal static Step GetFieldProvisioningStep(this Field templateField, TokenParser parser)
         {
-            XElement schemaElement;
-            if (!_fieldXmlDictionary.TryGetValue(templateField, out schemaElement))
-            {
-                schemaElement = XElement.Parse(parser.ParseXmlString(templateField.SchemaXml));
-                _fieldXmlDictionary[templateField] = schemaElement;
-            }
+            var schemaElement = GetCachedSchemaXml(templateField, parser);
             var type = (string)schemaElement.Attribute("Type");
             if (type != "Lookup" && type != "LookupMulti" && type != "Calculated")
             {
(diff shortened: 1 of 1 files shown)


gautamdsheth:
Thanks @svermaak , much appreciated help fixing this issue ,really awesome to have you contributing !!

PR #1261, opened 2026-07-31
Keep ListInstance DataSource across a template round trip

### LINKED ISSUE #1260: [BUG] ListInstance DataSource is dropped when a template is saved, and DefaultColumnValues are read as a DataSource
### Reporting an Issue or Missing Feature

Bug.

### Expected behavior

Reading a template and writing it back out keeps `<pnp:DataSource>` on an external list, and a list that declares no `DataSource` does not gain one.

### Actual behavior

On the current schema (2021-03 and 2022-09) `ListInstance.DataSource` is lost as soon as the template is written back, and a list that declares `<pnp:DefaultColumnValues>` but no `<pnp:DataSource>` comes back with the default column values sitting in `DataSource` as well.

Two separate causes, both in `Provisioning/Providers/Xml/Serializers/V202103/ListInstancesSerializer.cs`, which is the serializer selected for schema 2021-03 and later:

1. **Not written.** `Serialize` builds `dataSourceItemType` and its key/value selectors, then never registers the resolver. `V201705` and `V201909` both have `resolvers.Add($"{listInstanceType}.DataSource", ...)`; that line was not carried over when the 2021-03 serializer was added, so the selectors are dead code and the schema property is never populated.

2. **Read from the wrong property.** `DataSource` and `DefaultColumnValues` are both `StringDictionaryItem[]` on the schema list type, and neither resolver names its source property. When a list has no `DataSource`, `FromArrayToDictionaryValueResolver` falls back to "the first array property whose item type matches", finds `DefaultColumnValues` and reads that instead.

Round trip this template with `XMLPnPSchemaFormatter.LatestFormatter`:

```xml
<pnp:Provisioning xmlns:pnp="http://schemas.dev.office.com/PnP/2022/09/ProvisioningSchema">
  <pnp:Templates ID="CONTAINER">
    <pnp:ProvisioningTemplate ID="SAMPLE" Version="1">
      <pnp:Lists>
        <pnp:ListInstance Title="Customers" Url="Lists/Customers" TemplateType="600">
          <pnp:DataSource>
            <pnp:DataSourceItem Key="LobSystemInstance" Value="ODS" />
            <pnp:DataSourceItem Key="Entity" Value="ODS.Customer" />
          </pnp:DataSource>
        </pnp:ListInstance>
        <pnp:ListInstance Title="Documents" Url="Shared Documents" TemplateType="101">
          <pnp:DefaultColumnValues>
            <pnp:DefaultColumnValue Key="Field1" Value="Custom Value 1" />
          </pnp:DefaultColumnValues>
        </pnp:ListInstance>
      </pnp:Lists>
    </pnp:ProvisioningTemplate>
  </pnp:Templates>
</pnp:Provisioning>
```

`Customers` loads with its `DataSource` entries and comes back from the save with none. `Documents` loads with its `DefaultColumnValues` in `DataSource` as well.

The repo's own `Resources/Templates/ProvisioningSchema-2022-09-FullSample-01.xml` shows the same thing: its `Sample BCS List` loses the four `DataSourceItem` entries on a read/write round trip.

In PnP PowerShell terms, `Read-PnPSiteTemplate` followed by `Save-PnPSiteTemplate` drops the BCS configuration of any external list in the template.

### What is the version of the library you are running?

`dev` at 8a5506c, and the same code is in 1.19.x.

### Which operating system/environment are you running on?

Windows, .NET 9 and .NET 10, no tenant involved, the round trip is offline.

A PR follows.

svermaak:
## Type of change

- [x] Bug fix

## Related issue

Fixes #1260

Two problems in `Serializers/V202103/ListInstancesSerializer.cs`, the serializer selected for schema 2021-03 and 2022-09, so for anything current.

**1. `DataSource` was never written.** `Serialize` builds the item type and the key and value selectors and then never registers the resolver:

```csharp
// DataSource
var dataSourceItemType = Type.GetType(dataSourceItemTypeName, true);
var dataSourceItemKeySelector = CreateSelectorLambda(dataSourceItemType, "Key");
var dataSourceItemValueSelector = CreateSelectorLambda(dataSourceItemType, "Value");
                                             // <- no resolvers.Add, the three above are dead code

// Default Column Values
...
resolvers.Add($"{listInstanceType}.DefaultColumnValues", ...);
```

`V201705` and `V201909` both have `resolvers.Add($"{listInstanceType}.DataSource", ...)`. It was not carried over when the 2021-03 serializer was added, so reading a template and writing it back drops the BCS configuration of every external list in it. `Read-PnPSiteTemplate` followed by `Save-PnPSiteTemplate` is enough to lose it.

**2. `DefaultColumnValues` was read as a `DataSource`.** Both are `StringDictionaryItem[]` on the schema list type and neither resolver named its source property. When a list has no `DataSource`, `FromArrayToDictionaryValueResolver` falls back to "the first array property whose item type matches" and finds `DefaultColumnValues`:

```csharp
sourceValue = source.GetType().GetProperties().FirstOrDefault(sp => sp.PropertyType.IsArray &&
    sp.PropertyType.GetElementType().FullName == _sourceArrayItemType.FullName)?.GetValue(source);
```

Both resolvers now pass their source property name, which is the parameter that fallback already supports and that the `DataRows` resolver in the same file uses.

The two defects had been masking each other. Restoring the write path alone would have started writing the leaked values out as a `DataSource` the template never declared, so both belong in one change.

### Verification

Round trip of a template with one external list carrying four `DataSourceItem` entries and one library carrying `DefaultColumnValues` and no `DataSource`, through `XMLPnPSchemaFormatter.LatestFormatter`.

Before, the external list loses its BCS settings on save and the library picks up a `DataSource` it never declared:

Before the fix

After, the settings survive and stay where they belong:

After the fix

The repo's own `Resources/Templates/ProvisioningSchema-2022-09-FullSample-01.xml` behaves the same way: before this change its `Sample BCS List` came back from a round trip with no `DataSource`, and its `Projects` list came back with the default column values duplicated into `DataSource`.

Builds clean for `netstandard2.0`, `net8.0`, `net9.0` and `net10.0`.

### Regression test

`ListInstanceDataSourceTests` covers the three cases: `DataSource` survives a round trip, `DefaultColumnValues` survive a round trip, and `DefaultColumnValues` are not read as a `DataSource`. On `dev` the first and third fail, with this change all three pass.

The wider provider suite is unchanged: 451 passed, 15 failed before and after, the 15 being the `XMLProvidersTests` and `BaseTemplateTests` classes that need tenant details in `App.config`.

No CHANGELOG entry included.

### DIFF
--- a/src/lib/PnP.Framework.Test/Framework/Providers/ListInstanceDataSourceTests.cs
+++ b/src/lib/PnP.Framework.Test/Framework/Providers/ListInstanceDataSourceTests.cs
@@ -1,0 +1,87 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using PnP.Framework.Provisioning.Model;
+using PnP.Framework.Provisioning.Providers.Xml;
+using System.IO;
+using System.Linq;
+using System.Text;
+
+namespace PnP.Framework.Test.Framework.Providers
+{
+    [TestClass]
+    public class ListInstanceDataSourceTests
+    {
+        /// <summary>
+        /// One external list carrying BCS settings and one library carrying default column values.
+        /// Both settings are StringDictionaryItem arrays on the same list type, which is what the
+        /// serializers used to confuse.
+        /// </summary>
+        private const string Template = @"<?xml version=""1.0"" encoding=""utf-8""?>
+<pnp:Provisioning xmlns:pnp=""http://schemas.dev.office.com/PnP/2022/09/ProvisioningSchema"">
+  <pnp:Templates ID=""CONTAINER"">
+    <pnp:ProvisioningTemplate ID=""SAMPLE"" Version=""1"">
+      <pnp:Lists>
+        <pnp:ListInstance Title=""Customers"" Url=""Lists/Customers"" TemplateType=""600"">
+          <pnp:DataSource>
+            <pnp:DataSourceItem Key=""LobSystemInstance"" Value=""ODS"" />
+            <pnp:DataSourceItem Key=""EntityNamespace"" Value=""https://sample.contoso.com/ODS"" />
+            <pnp:DataSourceItem Key=""Entity"" Value=""ODS.Customer"" />
+            <pnp:DataSourceItem Key=""SpecificFinder"" Value=""CustomerRead Item"" />
+          </pnp:DataSource>
+        </pnp:ListInstance>
+        <pnp:ListInstance Title=""Documents"" Url=""Shared Documents"" TemplateType=""101"">
+          <pnp:DefaultColumnValues>
+            <pnp:DefaultColumnValue Key=""Field1"" Value=""Custom Value 1"" />
+            <pnp:DefaultColumnValue Key=""Field2"" Value=""Custom Value 2"" />
+          </pnp:DefaultColumnValues>
+        </pnp:ListInstance>
+      </pnp:Lists>
+    </pnp:ProvisioningTemplate>
+  </pnp:Templates>
+</pnp:Provisioning>";
+
+        [TestMethod]
+        public void DataSourceSurvivesATemplateRoundTrip()
+        {
+            ListInstance externalList = RoundTrip().Lists.First(l => l.Title == "Customers");
+
+            Assert.AreEqual(4, externalList.DataSource.Count, "the external list lost its DataSource entries");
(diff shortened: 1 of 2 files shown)


gautamdsheth:
Thanks @svermaak , much appreciated help fixing this issue !!

Issue #1248, opened 2026-07-05
Template provisioning can't handle section layouts or published state?


Magnus-K-work:
Found in: PnP.Framwork 1.18.0
Still exists in: PnP.Framework 1.19.46-nightly

How to recreate:
1. In a standard Teams site, create a news post from e.g. the template "Visual topic introduction" (A health announcement for fictional company Lenna)
2. Publish the Health news page
3. Using the template provisioning, copy the Health news to another site
4. In the target site a few bugs can be found:
  - The Health new page cannot be edited, in the developers console there are erros from trying to start the JS editors (PnP.Framework 1.18.0)
  - The Health news page is not published (PnP.Framework 1.19.0 and later)
  - The section layouts of the destination page are not consistant with the source news page:

Source layout:
<img width="1260" height="1088" alt="Image" src="https://github.com/user-attachments/assets/7b8a2528-3b89-40db-a008-321c723a7175" />

Destination layout:
<img width="1244" height="1082" alt="Image" src="https://github.com/user-attachments/assets/619e6c0f-0dde-4bba-92bc-7aae636a64fd" />
<img width="1263" height="1095" alt="Image" src="https://github.com/user-attachments/assets/cc802063-988a-4f7d-8c39-3f52dc1e049f" />
<img width="1222" height="1070" alt="Image" src="https://github.com/user-attachments/assets/55bbacc2-42fd-4b45-907c-48818590f61e" />
<img width="1245" height="1090" alt="Image" src="https://github.com/user-attachments/assets/0879c86c-05e8-4d4b-ab7e-316e285399da" />

Code used for copying the news page:
```
using Microsoft.SharePoint.Client;
using PnP.Framework;
using PnP.Framework.Provisioning.Model;
using PnP.Framework.Provisioning.ObjectHandlers;

/*
   .Net 8, Visual Studio 2026
   Nuget packages used:
   PnP.Framework 1.18.0, 1.19.0 and 1.19.46-nightly
 */

namespace HandlePageTemplate
{
    internal class Program
    {
        static void Main(string[] args)
        {
            // Credentials
            var appId = "<ID for app registration>";
            var certPassword = "<Password for certificate>";
            var certPath = "<Path to certificate for app registration>";
            var tenantName = "<tenant name>.onmicrosoft.com";

            // SharePoint Online Site URL
            var pageName = "<News post>.aspx";
            var sourceSiteUrl = "https://<tenant name>.sharepoint.com/sites/source_site";
[...]

PR #1271, opened 2026-08-18
Fix for 1248 - Corrected handling of section layout FlowLayout/Vertical in provisioning

Magnus-K-work:
Related issue:
#1248 - Template provisioning can't handle section layouts or published state?

What is changed:
- ClientSidePageContentsHelper.cs - lines 269-273 Uses correct layout type for FlexibleLayoutSection and FlexibleLayoutVerticalSection
- ClientSidePageContentsHelper.cs - line 203, re-throw correctly to keep stacktrace and inner exceptions
- TemplateListInstancesWriter.cs - line 127-129, handle null value for xmlField.Attribute("Required"), as this was causing null point exceptions in test runs

Tests:
- Not run in full, as many provisioning tests require installation of O365 Starter Intranet which is outdated/archived

Second public PR ever, please help me improve. :)

BR
Magnus


Magnus-K-work:
I'll keep these PRs alive by merging up from PnP.Framework / Dev for each release.

Adam-it (changes requested):
@Magnus-K-work awesome work so far, unfortunately the change is not complete as 
none of the provisioning XSDs or their generated classes have a flexible value in the schema CanvasSectionType enum.
When writing, `PnPObjectsMapper` converts enums with `Enum.Parse(dp.PropertyType, sourceValue.ToString())`. `Enum.Parse(schema.CanvasSectionType, "FlexibleLayoutSection")` will throw ArgumentException.

